PR #5, opened 2019-06-25
Replace SpreadsheetView with regular TableView

wirew0rm:
Replaces controlsfx SpreadsheetView with (open)JavaFX TableView because of controlsfx/controlsfx#1141.
Additionally contains some bugfixes and cleanup of Dataset Features and experimental FloatDataSet support.

Editing functionality is still in TableViewer, because it is difficult to factor out.
Also editing still throws some exceptions on editCommit, some task should propably be run via FXrun.

For the dataSets Travis complains about JavaDoc.


wirew0rm:
Heap usage (limited to -Xmx400M) while scrolling in table with 1M rows (ErrorDataSetRendererSample).
image

Issue #16, opened 2019-07-25
The second y axis cannot zoom in/out


soloman817:
so in example: https://github.com/GSI-CS-CO/chart-fx/blob/master/chartfx-samples/src/main/java/de/gsi/chart/samples/MultipleAxesSample.java

If I lock the zoomed to only zoom Y axis, and then scroll the mouse, only the left y-axis (the first y axis) zooms, the 2 other y axis doesn't change. 

Is there way to select which y-axis to zoom?

RalphSteinhagen:
Not yet. This is a missing functionality in the Zoomer plugin that for the time being tracks only changes of the first primary axes for X and Y.

https://github.com/GSI-CS-CO/chart-fx/blob/a67b98c8224bdaccb627b60c123e4570cff1b17b/chartfx-chart/src/main/java/de/gsi/chart/plugins/Zoomer.java#L821-L832

The functionality seems to be limited to above function (+ ZoomState tracking) and would need to be extended to 
   * loop through all axes, 
   * track changes and performs the zoom for each one, and
  * ideally with a boolean flag that controls whether the axis is taken into account.

The first two items should be easy to do and do not require API changes (ie. zooming always for all axes). The boolean flag for specific axes may need a Zoomer plugin API extension. Should be robust enough to handle axes being added/removed during intermediate zoom states.

The MutipleAxesSample is sufficient for testing/documenting the new functionality.

RalphSteinhagen:
some addition: whether a axis should be taken into account for zooming or not could be controlled with the following marks:

https://github.com/GSI-CS-CO/chart-fx/blob/a67b98c8224bdaccb627b60c123e4570cff1b17b/chartfx-chart/src/main/java/de/gsi/chart/plugins/Zoomer.java#L971-L975

I presume that one would need one ZoomState object per Axis, ie.

https://github.com/GSI-CS-CO/chart-fx/blob/a67b98c8224bdaccb627b60c123e4570cff1b17b/chartfx-chart/src/main/java/de/gsi/chart/plugins/Zoomer.java#L123

being a ArrayDeque of a List of ZoomState (one per y-axis).

RalphSteinhagen:
@soloman817 if you like to check the prototype implementation. We may merge this with the master/JDK production branch by next week.
https://github.com/GSI-CS-CO/chart-fx/tree/feature-zoomer-extension

soloman817:
[...]

PR #20, opened 2019-08-06
Change Zoomer plugin to zoom all axes

RalphSteinhagen:
All axes are zoomed and can be selectively excluded from zooming.

resolves user issue #16


wirew0rm:
I have no idea why travis is failing here... other PRs are fine, master is fine and the diff have nothing relevant to CI...

RalphSteinhagen:
> Also how does Zoomer handle non-numeric axes now? Are they ignored or zoomed now? Just since the comment restricting the use was removed.

with one of the earlier refactoring all axes (including e.g. CategoryAxis) map to a Axis. The NumericAxis was a legacy from older times where there were multiple base-classes for axes... not needed anymore.

PR #27, opened 2019-09-11
removed duplicate plugin

ennerf:
Currently `mvn package` fails. Tested with mvn 3.6.0 / 3.6.2 and JDK 8 / 11 / 12

image

This is caused by a limitation in maven that doesn't allow for the same plugin to be added more than once. This PR combines both `<plugin>` definitions into separate `<execution>` elements.

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -121,6 +121,13 @@
                             </goals>
                             <phase>site</phase>
                         </execution>
+			<execution>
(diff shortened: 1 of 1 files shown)


ennerf:
Sorry, you're right. I didn't see that there was another `<pluginManagement>` section. I've never seen maven simply break like this, but in this case the issue has to be somewhere else.

PR #63, opened 2019-12-11
readLock deadlocks in DataSetUtils

graynk:
So, I noticed that when I try to serialize datasets using `DataSetUtils.writeDataSetToFile()` it deadlocks on `dataSet.lock().readLockGuard()` in `writeDataSetToByteArray()`. I'm guessing it's supposed to be `writeLock`, not `readLock`, similar to [this](https://github.com/GSI-CS-CO/chart-fx/commit/df31ebde356860db4e09925a9847b808eb459f74) commit, only in reverse. At least it doesn't deadlock for me anymore.

### DIFF
--- a/chartfx-dataset/src/main/java/de/gsi/dataset/utils/DataSetUtils.java
+++ b/chartfx-dataset/src/main/java/de/gsi/dataset/utils/DataSetUtils.java
@@ -838,7 +838,7 @@ public class DataSetUtils extends DataSetUtilsHelper {
         }
 
         byteOutput.reset();
-        dataSet.lock().readLockGuard(() -> {
+        dataSet.lock().writeLockGuard(() -> {
             try {
                 byteOutput.write(("#file producer : " + DataSetUtils.class.getCanonicalName() + '\n').getBytes());
 


wirew0rm:
Hey there, thanks for using our code and also for contributing back :)

Since serializing a dataSet should not require a write lock, the issue must be somewhere else, so more information on how you call the function (which DataSet, what happens on other threads (if you have multiple threads) and which locks do they hold/which dataset concerned tasks do they do) would be useful for us to understand where the real problem is here.

Are you calling `writeDataSetToFile()` while you are holding a WriteLock in your user code? While it is theoretically possible to downgrade a WriteLock to a ReadLock, this is not implemented in our version, see the last bullet point in [ThreadSafety.md](https://github.com/GSI-CS-CO/chart-fx/blob/master/docs/ThreadSafety.md).

Can you provide a minimal example of what you are trying to do? Or even better a corresponding test case to [DataSetUtilsTest](https://github.com/GSI-CS-CO/chart-fx/blob/master/chartfx-dataset/src/test/java/de/gsi/dataset/utils/DataSetUtilsTest.java)?

While your proposed fix might fix your deadlock issue, it blocks alll other access to the dataSet while doing potentially expensive IO operations.

graynk:
> Since serializing a dataSet should not require a write lock, the issue must be somewhere else, so more information on how you call the function (which DataSet, what happens on other threads (if you have multiple threads) and which locks do they hold/which dataset concerned tasks do they do) would be useful for us to understand where the real problem is here.

I really should have created an issue first, I just figured that it was really similar to another blocking issue that I've had before, sorry about that.

> Can you provide a minimal example of what you are trying to do? Or even better a corresponding test case to [DataSetUtilsTest](https://github.com/GSI-CS-CO/chart-fx/blob/master/chartfx-dataset/src/test/java/de/gsi/dataset/utils/DataSetUtilsTest.java)?

Yesterday I was able to reproduce locking stably, time after time. Today I can't reproduce it for the life of me. Sorry, I'll close this for now and create an issue once I have more information about this (if this even needs fixing).

PR #67, opened 2019-12-18
Use first dataset name in measurement plugin when no choice is required

dedeibel:
If only one dataset was available, no name was used instead of using the first one.

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/AbstractChartMeasurement.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/AbstractChartMeasurement.java
@@ -115,8 +115,12 @@ public abstract class AbstractChartMeasurement implements EventListener {
         }
 
         if (dataSet == null) {
-            valueField.setDataSetName("<unknown data set>");
-            this.dataSet = dataSet;
+            if (dataSetSelector.getNumberDataSets() == 1) {
+                valueField.setDataSetName(dataSetSelector.getFirstDataSet().getName());
+            } else {
+                valueField.setDataSetName("<unknown data set>");
+                this.dataSet = dataSet;
+            }
         } else {
             valueField.setDataSetName("<" + dataSet.getName() + ">");
             this.dataSet = dataSet;
--- a/chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/utils/DataSetSelector.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/utils/DataSetSelector.java
@@ -41,6 +41,10 @@ public class DataSetSelector extends HBox {
     public DataSet getSelectedDataSet() {
         return dataSets.getSelectionModel().getSelectedItem();
     }
+    
+    public DataSet getFirstDataSet() {
+        return dataSets.getItems().get(0);
+    }
 
     static protected class DataSetLabel extends ListCell<DataSet> {
         @Override


chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/AbstractChartMeasurement.java
```diff
@@ -115,8 +115,12 @@ public void setDataSet(final DataSet dataSet) {
...
 
         if (dataSet == null) {
-            valueField.setDataSetName("<unknown data set>");
-            this.dataSet = dataSet;
+            if (dataSetSelector.getNumberDataSets() == 1) {
+                valueField.setDataSetName(dataSetSelector.getFirstDataSet().getName());
+            } else {
+                valueField.setDataSetName("<unknown data set>");
+                this.dataSet = dataSet;
+            }
```
wirew0rm: This looks weird, from what I understand, this would set the name of the first data set, but leave the data set at its old value and it would also not setup the listener?
I suspect this only works for measurement plugins which do not really need the dataset anyway.
If the list of DataSets is not empty, dataSet should probably not be `null` here anyway imho.

(I noticed problems with DataSet selection in my App, too, but I haven't come around to look into it until now so thanks for bringing this up, i'll investigate)

wirew0rm: [These 3 lines](https://github.com/GSI-CS-CO/chart-fx/blob/master/chartfx-chart/src/main/java/de/gsi/chart/plugins/measurements/utils/DataSetSelector.java#L28-L30) seem to not have the intended effect.

dedeibel: They seem to work since there is a solution in place in this method:

```
public DataSet getDataSet() {
        if (dataSet == null) {
            final List<DataSet> allDataSets = new ArrayList<>(chart.getAllDatasets());
            return allDataSets.get(0);
        }
        return dataSet;
}
```

wirew0rm: Still does not add the listener and it is not obvious to me that this will always return the same DataSet as the one whose name is shown in the box esp when datasets are added and removed frequently.

(I also have not yet looked into what the measurement implementations do with this dataset, esp since a DataSet might even be drawn by different renderers using different axes. But this is probably out of scope for this PR.)

dedeibel: I do not understand the problem yet. I will think about it an reopen the PR when I have more insight.

wirew0rm:
I found the problem causing this at the root:
The DataSet selection uses the list returned by chart.getAllDataSets(). This function always returns the same ObservableArrayList, but on each call, the list is cleared and refilled. This happens all the time, because eg the axes use it to rescale to the data. When the list is cleared, the selection model is reset. 
```
 allDataSets = FXCollections.observableArrayList(chart.getAllDatasets());
```
fixes this issue, but then the listView only updates when the dialog is closed and reopened, but this is imho not too bad.
Now by default, the first DataSet is selected in the dialog and returned, making it unnecessary to introduce a special treatment when only having one dataset. (But maybe we could skip showing the dialog then completely, but again out of scope for now)

PR #69, opened 2019-12-19
fixed indentation in restyled bot clang formatter definition

RalphSteinhagen:


.restyled.yaml
```diff
@@ -178,26 +178,27 @@ restylers_version: "20191031"
...
-- "**/*.hh"
-- "**/*.hpp"
-- "**/*.hxx"
-- "**/*.h++"
-- "**/*.H"
-- "**/*.java"
-- "**/*.js"
-- "**/*.m"
-interpreters: []
+  - name: prettier
```
wirew0rm: name should be [clang-format](https://github.com/restyled-io/restyled.io/wiki/Available-Restylers#clang-format) not [prettier](https://github.com/restyled-io/restyled.io/wiki/Available-Restylers#prettier).
This probably works to, but uses the prettier javascript restyler but changes all it's default settings to the default settings of the clang-format restyler.

RalphSteinhagen: fixed in commit https://github.com/GSI-CS-CO/chart-fx/pull/69/commits/b7f80b4cef954ca52696fa2487eb9465e98bdd0f

PR #128, opened 2020-03-19
Change codacity coverage upload from flaky bash script to maven plugin

wirew0rm:
Also move owasp dependency check behind coverity, so that it does not block releases.


RalphSteinhagen:
lgtm :thumbsup:

PR #134, opened 2020-03-23
Add instructions to use snapshot repository

wirew0rm:


ennerf:
I've already mentioned this to @RalphSteinhagen, but I think Jitpack would be a better way to use unreleased versions. It (also) allows users to specify individual commits rather than only the latest release in a given branch.

I've relied on this a few times in the past when I needed a new feature or bugfix, but had to make sure that CI still works and won't suddenly break due to potentially unknown changes.

```xml
<repositories>
	<repository>
		<id>jitpack.io</id>
		<url>https://jitpack.io</url>
	</repository>
</repositories>

<dependencies>
	<dependency>
		<groupId>com.github.GSI-CS-CO.chart-fx</groupId>
		<artifactId>chartfx-chart</artifactId>
		<version>1c197fe</version> <!-- or e.g. master / 11.1.3-->
	</dependency>
</dependencies>
```

If you setup [a custom domain](https://jitpack.io/docs/#custom-domain-name) it would allow users to use the same maven coordinates as the official maven repo.

wirew0rm:
@RalphSteinhagen The rationale for putting it into the development section was, that it is not intended to be used by regular users, but only by users that have at least some understanding of the unreleased changes, e.g. for verifying bug fixes, but this can also be explained in the text.
Collapsing is a good idea, so i'll move it up and add it as a collapsed section.

@ennerf We don't really want to prominently recommend jitpack because we do not use it ourselves and it has it's own build process where we do not get notifications if something breaks, tests are not run, they might use a different jdk version, etc ...
I see your point about pinning specific commits, it is a bit unfortunate that maven uses the time and not the git hash to qualify SNAPSHOT releases, but you can pin them with `<version>master-20200320.180559-79</version>` for the current master.
Regarding a custom domain I think this would be difficult with the way our coordinates are currently chosen (there is no 1:1 mapping between repositories and coordinate prefixes) and once you had jitpack setup, it would depend on the ordering of repositories if you get releases from mavencentral or jitpack, so I believe it's also better to make this explicit if users want to use jitpack.

PR #147, opened 2020-04-06
Fix mvn build for local builds and external PRs

wirew0rm:
Remove execution phase for codacy upload so it is only triggered by running `mvn codacy:coverage` explicitly.
Only run coverage upload if codacy api token environment variable is set.
Add headless options to build command in configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
 
 after_success:
   # create and upload coverage report
-  - mvn jacoco:report-aggregate coveralls:report
+  - mvn jacoco:report-aggregate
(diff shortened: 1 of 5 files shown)


wirew0rm:
As suggested by @RalphSteinhagen also removed coveralls in favour of codecov and moved headless test configuration into pom.xml, so that unit tests allways run headless. Running tests from within eclipse still uses the non-headless mode.

Also enabled parallel test execution for chartfx-dataset. While this does not speed up travis builds significantly, it can make local builds faster and is a first step towards enabling parallel tests for the other modules. I also tried activating it for chartfx-math, but some performance samples there encountered timeouts. (This will probably need some https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/parallel/Execution.html Annotations to make it work.)

Issue #246, opened 2020-09-01
Histogram bin width does not match actual bin range


SamuelLKane:
Hi all! I'm trying to use ChartFX's Histogram to plot pre-binned data and noticed a discrepancy that is echoed in the sample code (HistogramSample.java). 

### What are the issues:

1. The edges of the bin do not reflect the actual range of the bin
2. The data point is not always the center of the bin, instead it can be another point within the bin

### Examples

![image](https://user-images.githubusercontent.com/16545929/91899864-c50f6e00-ec6b-11ea-9519-3baf8e8b1ee8.png)
This image shows the green dataset3 from HistogramSample.java which has fixed bins going from 3-4,4-5, and 5-10 (among others). In this case all the visible data points are correctly centered on the bins but the edges of the bins are incorrectly placed (this has the effect of making it look like the data points are incorrectly placed). This is most obvious with the left edge of the bin going from 5-10 which begins at 6, not 5.

![image](https://user-images.githubusercontent.com/16545929/91899816-af01ad80-ec6b-11ea-8d08-6c2e7e7317d2.png)
This image is from our own implementation using pre-binned data and shows the data points which are placed above the lower bound of a bin, resulting in a histogram that is off by 1/2 a bar. The visible bins are -0.5-10.3, 10.3-21.1, 21.1-31.9, 31.9-42.6.

RalphSteinhagen:
@SamuelLKane do you use by chance non-equidistant bins (see also discussions in issue #80)?

SamuelLKane:
@RalphSteinhagen Thanks for the reference. I've poured over issue 80 and confirmed we're using equidistant bins (all are 10.8 in size). We are using the new Histogram(title, numBins, min, max, horizontal) constructor that should create equidistant bins regardless and I confirmed it was via .isEquiDistant(). I swapped to the new Histogram(title, double[]) and was able to get a correctly binned and rendered histogram. I don't know if it's relevant but we also loose the first bin's data in the render when using the numBins, min, max constructor (it only displays bins 1 - numBins as if they were one bin less; bin 1's height in bin 0's place, bin 2's height in bin 1's place, etc.)

wirew0rm:
This is probably a classic of by 1 error, the convention for the Histogram class is that bin 0 and bin nBins+1 contain the under-/overflow data, so the actual data bins start at index 1.

This was only documented in `Histogram#getBinContent(int)`I've added a note to the interface itself.
[...]

Issue #237, opened 2020-08-22
Changing horizontal minor grid line visibility doesnt refresh chart


tanrikut:
Calling `xyChart.getGridRenderer().getHorizontalMinorGrid().setVisible(visibility);` doesnt update the chart visual whereas changing major axis visibility updates the chart `xyChart.getGridRenderer().getHorizontalMajorGrid().setVisible(visibility);`

wirew0rm:
solved on the development branch, will be merged and released soon, thanks for reporting :+1:

tanrikut:
Thanks

PR #249, opened 2020-09-07
Histogram, GridRenderer redraw and chartfx-generator with mvn clean

wirew0rm:
This PR adds a missing redraw listener (see #237) and adds more documentation, an additional sample and some small changes to Histogram to adress #246.
Also fixes the issue that mvn clean after mvn compile would fail if the chartfx-generate plugin was not in the local or remote repository, because it would be deleted before the modules using it. This was solved by adding the compile step to the clean step of the plugin.


RalphSteinhagen (approved):
LGTM

PR #258, opened 2020-09-11
Serialiser coverity fixes

wirew0rm:
Fixes two classes of issues in the Serialiser discovered by coverity:
- The IoSerialiser needs access to the registered FieldSerialisers
- The return functions are never used for primitve / boxed / array types and did not have null checks. fixed by throwing errors in the return functions, alternative: add null checks, should have no performance impact, because the code is actually never called. I prefer the exceptions, because it reduces the amount of code we have to think about, but if there are cornercases where the return functions are needed we should use the null checks.


RalphSteinhagen (approved):
Very good fixes and I like the improved IoSerialiser API! :+1:

Issue #261, opened 2020-09-22
What is the professional way to invert xAxis values from max to min?


osaidmakhalfeh:
Hi Dears,
First of all I want to thank you again for this amazing work!

Here, I'm trying to draw the xAxis from the "Max" to the "Min" values.
I'm not sure, how can I do it easily? 

![inxax](https://user-images.githubusercontent.com/44381968/93913699-1ca27600-fd0e-11ea-9741-5c2946b7eb46.png)

Thanks a lot

milo-gsi:
Hi osaidmakhalfeh,

this a property of the axis.

See invertAxis(boolean)

osaidmakhalfeh:
> Hi osaidmakhalfeh,
> 
> this is a property of the axis.
> 
> See invertAxis(boolean)
Thanks a lot for your response.
I've done that before, but unfortunately, the numbers have disappeared from xAxis, here an eg.
![image](https://user-images.githubusercontent.com/44381968/93923360-f71c6900-fd1b-11ea-8fc8-477e51acd8cf.png)

milo-gsi:
Hi osaidmakhalfeh,

we just found that issue as I wanted to point you to one of the samples.
I'm sure it'll be fixed soon or @RalphSteinhagen / @wirew0rm will provide a workaround.

RalphSteinhagen:
@osaidmakhalfeh can you give it a try with the fixed version in the 'fix261' branch?

If you do not want to compile yourself, you can also use the snapshot repository as described [here](https://github.com/GSI-CS-CO/chart-fx#using-the-snapshot-repository). Just enable the snapshot repository and use `fix261-SNAPSHOT`for the version.

RalphSteinhagen:
... a fix a day keeps harm a way. 

@osaidmakhalfeh thanks for asking a question and triggering us to fix that issue. This must-have slipped in during one of the last commits and we did not catch this earlier.

Let us know if this fixes and helps with your particular application

RalphSteinhagen:
... now merged to master. Thanks to @milo-gsi for cross-checking.

osaidmakhalfeh:
A lot of thanks, Guys.
Now it works!

But there are a new icons that appeared after the fix.
![image](https://user-images.githubusercontent.com/44381968/94005786-75721d00-fda7-11ea-9396-fe1ec981801f.png)

 I'm not sure what is that, will try to solve it!
Thanks again.

[...]

PR #262, opened 2020-09-22
fix261 -- fixes inverted x-axis problem

RalphSteinhagen:
fixes wrong label overlap detection that prevented inverted x-axes (see #261)


milo-gsi (approved):
LGTM

ErrorDataSetRendererStylingSample works for me now.

PR #269, opened 2020-09-30
Fix error data set renderer for inverted x axis

### LINKED ISSUE #265: BarChart: non equidistand data and last data point not drawn
Hi Dears,

I'm facing a problem while drawing a barChart with small values for the yAxis.

here is a snapshot for the values I'm trying to draw:

asas

Unfortunately, the result looks like that there is one bar rendered, but it took the full width of the screen, here is a snapshot of it.

floatissue

here is my sample to share the knowledge.
https://github.com/osaidmakhalfeh/testChartsProject/blob/master/pcCharts/src/main/java/BarChart.java

Thanks again.

wirew0rm:
For inverted x-Axis, the Renderer first checked for the points to
render and then swapped them and tried to account for the fact that the
points might not be exactly on the axis bounds by adding a fixed extra
point to each side, which should have been two, because the code which
determines which points to include for normal axes also adds one in the
other direction.
This commit replaces this by the cleaner approach of obtaining xMax at
the left bound of an inverted axis and vice versa.

Also adds a a switch for assumeSorted to the ErrorDataSetStylingSample.

fixes #265


RalphSteinhagen (approved):
well spotted.LGTM

PR #271, opened 2020-10-01
Fix bug in error point reduction

wirew0rm:
When reducing the points, the reducer would for some reason swap the
positive and negative error for the last point. When no actual reduction
is taking place, this leads to the values actually having the same value
which would give strange results in the error bars/surfaces.

20200929_datareduction_errors

This commit removes the swapping of the errors of the last point.

The error surface renderer would also swap these points again before
rendering, so this is also removed with this commit.

To ease testing, an option to enable/disable parallel point caching in
the renderer was added to the ErrorDataSetRendererStylingSample.


RalphSteinhagen (approved):
LGTM

PR #274, opened 2020-10-09
Exclude tests from coverage

wirew0rm:
This PR excludes tests, benchmarks and samples from the coverage computation.

Additionally adds the microservice tot the reports submodule where it was missing.


RalphSteinhagen (approved):
thanks. LGTM-

PR #281, opened 2020-10-21
fixes IoClassSerialiserTests for custom pojos directly serialised as rootObj

RalphSteinhagen:
The corresponding routine in IoSerialiser
<E> E getCustomData(FieldSerialiser<E> serialiser);
does not provide any provisions to make in-place modification and thus always returns a new custom object. 

Thus the return value needs to be always used for testing the deserialisation identity. Updated the test accordingly.

N.B. this is different for custom objects that are nested within another (generic) pojo. In this case the regular return/read FieldSerialiser call-back functions are used.


wirew0rm (approved):
Looks good. :+1: 
We could think about adding `@CheckReturnValue` annotations (equivalent to c++ `[nodiscard]`, but only evaluated by e.g. findbugs, not the compiler) to the `deserialiseObject` functions to try to prevent this class of bugs.

PR #285, opened 2020-10-23
Enable rerunning failed unit tests

wirew0rm:
This is needed because on the travis runner there are some non
deterministic test failures for the allocator based tests and the testfx
based ones.

Because this feature needs feature needs surefire > 3.0.0-M4 I updated
all build plugins to their latest release.


RalphSteinhagen (approved):
LGTM! :+1:

PR #287, opened 2020-10-27
Add scroll zooom on individual axis

wirew0rm:
This is a work in progress to allow zooming on individual axes.

- [x] scroll zoom on individual axes
  - [ ] fix wrong bounding box for chart area, zooms chart instead of axis for upper half of the axis
- [ ] left drag to zoom axis
- [ ] middle drag to pan axis
- [ ] test cases in ZoomerTest

Right click on axes to undo zooms is not possible because it is already used for the `EditAxisPlugin`, but right clicking on the chart is sufficient.


wirew0rm:
With the current chart layout scenegraph topology and plugin event listener registration infrastructure it is not easily possible to implement this consistently, so I'm going to close this for now.
Once we did the simplification of the layout we can start a new attempt at implementing this feature.

Issue #286, opened 2020-10-27
DataPointTooltip issues


magicroomy:
First of all your ChartsFX is phantastic. In the past I used JFreeChart due to lack of alternatives. ChartFX is a killer.
I try to add tooltips to a multi axis line chart. I do as found in the examples:

.... chart.getPlugins().add(new DatapointTooltip());

1) I noticed that tooltips are only shown for the "primary" dataset/axis specified when creating the chart. I would like to have the tooltips for all points in all datasets/axis. Any hint on that for me?
2) While AbstractDataFormattingPlugin has methods for setXValueFormatter and setYValueFormatter, looking into the code for DatapointTooltip reveals, that this class completely ignores those formatters. Is this intentional or do I miss something?

wirew0rm:
Hey there and thanks for you interest in chart-fx.

1) This is a known shortcoming of the plugin, there is even a [small note in the source](https://github.com/GSI-CS-CO/chart-fx/blob/master/chartfx-chart/src/main/java/de/gsi/chart/plugins/DataPointTooltip.java#L105).(It also only supports sorted x/y datasets properly.) If you want to give it a try to implement this, feel free to ask your questions here and/or submit a draft PR with your progress.

2) I aggree it would probably be cleaner to use the already existing interface I don't know about the origin of this choice. In the meantime there is a workaround [here](https://github.com/GSI-CS-CO/chart-fx/issues/268#issuecomment-699672130).

Let me know if you have any further questions.

wirew0rm:
I implemented the things outlined in my last message. Could you test them by using the development version `akrimmTooltipForAllRenderers-SNAPSHOT`? ([instructions](https://github.com/GSI-CS-CO/chart-fx#using-the-snapshot-repository) here, should be available as soon as travis processed the pull request, alternatively you could of course build the branch yourself).

magicroomy:
> 
> 
> I implemented the things outlined in my last message. Could you test them by using the development version `akrimmTooltipForAllRenderers-SNAPSHOT`? ([instructions](https://github.com/GSI-CS-CO/chart-fx#using-the-snapshot-repository) here, should be available as soon as travis processed the pull request, alternatively you could of course build the branch yourself).

Hi,
I checked it out and gave it a first check.

So far it looks good.
Tooltip shown on all curves/Axis + DataPointTooltip accepts formatters.

Excellent Work, thanx a lot. Looking forwar to the next "Major" Release

[...]

PR #293, opened 2020-10-30
Show Tooltips for all renderers and dataset in the chart and use the either the specified formater or the axis formater

wirew0rm:
* Allow the Tooltip to be shown for all renderers and datasets shown in the chart
* use the formaters which are customizable in the AbstractDataFormating parent plugin or if they are not available the formaters of the axis
* Add TestFx tests for the tooltip

solves #286


RalphSteinhagen (approved):
LGTM!

PR #296, opened 2020-11-02
Allow serialisation of empty subelements in cmwLightSerialiser

wirew0rm:
Add a testcase with a message serialized by the reference serialiser and fix the bug it triggers.
Also changes the auto protocol detection mechanism because before it relied on the size field of cmw never to be zero, now use negative field size instead.


RalphSteinhagen (approved):
Sehr gut!

RalphSteinhagen (approved):
Still very good!

PR #298, opened 2020-11-03
Disruptor demux aggregation proof of concept

wirew0rm:
Demonstrates aggregation of multiple overlapping event sequences into separate aggregation workers and republishing of aggregate events.
Use an event generator to test different event scenarios.


RalphSteinhagen (approved):
Very nice concept! LGTM

PR #299, opened 2020-11-03
Coverity fixes

RalphSteinhagen:
nomen est omen see commits for details

masked two false-positive coverity issues, related to:

try {
 // [..]
} catch (final NoSuchAlgorithmException | CertificateException | KeyStoreException | IOException e) {
  LOGGER.atError().setCause(e).addArgument(keyStoreFile == null ? "internal" : keyStoreFile).log("error while reading key store from '{}'");
   	
CID 303804 (#1 of 1): Unlogged security exception (UNLOGGED_SECURITY_EXCEPTION)
9. end_of_catch: Arriving at the end of a catch block without logging the security exception.
}


wirew0rm (approved):
:+1:

PR #302, opened 2020-11-05
Fix scroll zoom in

### LINKED ISSUE #301: Scroll zoom only works for zooming out
When using the zoomer, it is only possible to zoom out, not to zoom in. This used to work in earlier versions, so we'll have to track what changed.

milo-gsi:
Local mouse scroll events send additional zeroes.
Discard them.

fixes #301


RalphSteinhagen (approved):
Well spotted! Thanks for the qucik fix. LGTM!

PR #304, opened 2020-11-06
Fix memory leak in CssPropertyFactory

wirew0rm:
Remove/weaken two static references to instantiated objects in the CssPropertyFactory.
Also adds a simple test which validates, that after removing a chart from the scene at least the big components will be garbage collected.


dedeibel:
LGTM thank you

dedeibel (approved):
LGTM

Issue #303, opened 2020-11-06
Test and Javadoc errors with openjdk14 on windows


ArjenSturm:
I like to add a candlestick renderer for candlestick charts for cycle analysis, but I get the errormessage below for several test classes when I open and build the cloned chart-fx source in my Apache Netbeans 12 IDE as Maven project, so I can't now build and run some samples.
It looks like this issue : http://mail.openjdk.java.net/pipermail/openjfx-dev/2018-April/021735.html
And that has to do with running TestFX in headless mode via Monocle and screen size.
I have attached the build output log.

[build_output.log](https://github.com/GSI-CS-CO/chart-fx/files/5501214/build_output.log)
```
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse integer for 'malformedInt'='0.ag' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse floating point for 'malformedInt'='0.ag' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse floating point for 'emptyProperty'='' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse floating point for 'floatingPointArray'='' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse floating point for 'malformedInt'='0.ag' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse color description for 'color'='reddish' returning null
[main] ERROR de.gsi.chart.utils.StyleParser - could not parse color description for 'malformedInt'='0.ag' returning null
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.014 s - in de.gsi.chart.utils.StyleParserTest
Running de.gsi.chart.plugins.measurements.utils.DataSetSelectorTests
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.656 s - in de.gsi.chart.plugins.measurements.utils.DataSetSelectorTests
Running de.gsi.chart.plugins.CrosshairIndicatorTest
java.nio.BufferOverflowException
	at java.base/java.nio.DirectIntBufferU.put(DirectIntBufferU.java:457)
	at com.sun.javafx.tk.quantum.UploadingPainter.run(UploadingPainter.java:162)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
[...]

PR #306, opened 2020-11-06
Reduce scene size for headless tests

wirew0rm:
Reduce scene sizes for testfx tests because monocle has a frame buffer size limitation.

see #303


wirew0rm:
hm :confused:  The BufferOverflowException still occurs:
https://travis-ci.org/github/GSI-CS-CO/chart-fx/builds/741887210#L2996-L3026

RalphSteinhagen (approved):
:+1: benign

PR #323, opened 2020-11-27
Correct automatic module names

### LINKED ISSUE #321: Hyphen module name breaks module based projects
As per the JSE 9+ specs;

https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html#jls-ModuleDeclaration
https://docs.oracle.com/javase/specs/jls/se9/html/jls-3.html#jls-3.8

A module/package identifier, within the module descriptor(module-info.java) needs to be an alphanumeric string. The suggested workaround for module based projects is to use a period in place of the hyphen, for the "Automatic-Module-Name" in the manifest as such:

Automatic-Module-Name: de.gsi.chart.chartfx.chart

Further info on module resolution:

https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleFinder.html

Love your work.

wirew0rm:
Module names should not contain hyphens, so we cannot reuse our artifact
names. Instead introduce a new property which will be set in each
module.

fixes #321

### DIFF
--- a/chartfx-acc/pom.xml
+++ b/chartfx-acc/pom.xml
@@ -11,6 +11,9 @@
     <groupId>de.gsi.acc</groupId>
     <artifactId>chartfx-acc</artifactId>
     <name>acc</name>
+    <properties>
+        <project.moduleName>de.gsi.chartfx.acc</project.moduleName>
+    </properties>
 
     <description>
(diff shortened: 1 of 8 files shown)


RalphSteinhagen (approved):
:+1: addresses issue #321

Issue #337, opened 2021-01-15
DataPointTooltip missing dataset synchronization


dedeibel:
I think there might be race condition in the `DataPointTooltip` code regarding the lookup of points near the mouse and getting the data set's values.

First the index of valid points is fetched, then later it's value is fetched. In the meantime the dataset could have been cleared resulting in an out of bounds exception.

https://github.com/GSI-CS-CO/chart-fx/blob/fc81f1fa2fced4d4102bbc4fa3558cef35742a7e/chartfx-chart/src/main/java/de/gsi/chart/plugins/DataPointTooltip.java#L126

And then:

https://github.com/GSI-CS-CO/chart-fx/blob/fc81f1fa2fced4d4102bbc4fa3558cef35742a7e/chartfx-chart/src/main/java/de/gsi/chart/plugins/DataPointTooltip.java#L134

I observed the following stacktrace regarding this:

```
StackTrace
    java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
    	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
    	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
    	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
    	at java.base/java.util.Objects.checkIndex(Objects.java:372)
    	at java.base/java.util.ArrayList.get(ArrayList.java:459)
    	at de.gsi.dataset.spi.LabelledMarkerDataSet.get(LabelledMarkerDataSet.java:74)
    	at de.gsi.chart.plugins.DataPointTooltip.getDataPointFromDataSet(DataPointTooltip.java:134)
    	at de.gsi.chart.plugins.DataPointTooltip.lambda$getPointsCloseToCursor$5(DataPointTooltip.java:129)
    	at java.base/java.util.stream.IntPipeline$1$1.accept(IntPipeline.java:180)
    	at java.base/java.util.stream.Streams$RangeIntSpliterator.forEachRemaining(Streams.java:104)
    	at java.base/java.util.Spliterator$OfInt.forEachRemaining(Spliterator.java:699)
    	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
    	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
[...]

PR #339, opened 2021-01-26
Fix337 tooltip synchronization

dedeibel:
I tried to come up with a fix to #337

As suggested by Alex I tried to calculate all DataPoint information inside a read lock. One for each data set to keep it as small as possible.


wirew0rm (approved):
Looks good to me. I also like the readability improvements, thanks a lot. I'll merge this later into master later today if there are no objections.

PR #280, opened 2020-10-21
Add table view only to chart if really shown

dedeibel:
- Mitigates problem of it taking control over the charts size

I talked about it with akrimm, please ask if you want more info.

PR #341, opened 2021-01-26
TableViewer: table cells can now be selected again

dedeibel:
TableViewer: table cells can now be selected again

- Mouse transparency was set incorrectly after commit eb7a286 (#280)

Added a test for the specific problem.

(sorry)


wirew0rm (approved):
LGTM :+1:

PR #343, opened 2021-01-26
TableViewer: fix calls to hashCode of DataSetsRow

dedeibel:
While working on a bugfix for the `TableViewer` I noticed a problem when the `DataSetsRow` was added to a Hashing collection. (For example the result of a `fxRobot` lookup, resulting in a Set).

The implementation of the `DataSetsRow` `equals`/`hashCode` seem to use to the `model`\`s `equals`/`hashCode` which in turn uses the Lists default implementations. Those will in turn call it's children's methods, being the `DataSetsRows`.

Try using the original methods to see a stack overflow in the test case.

I am not sure about the best implementation of the `equals`/`hashCode` methods in this case, please see it just as a suggestion.

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/plugins/TableViewer.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/plugins/TableViewer.java
@@ -742,7 +742,7 @@ public class TableViewer extends ChartPlugin {
         public boolean equals(final Object o) {
             if (o instanceof DataSetsRow) {
                 final DataSetsRow dsr = (DataSetsRow) o;
-                return ((dsr.getRow() == row) && model.equals(dsr.getModel()));
+                return ((dsr.getRow() == row) && model == dsr.getModel());
             }
             return false;
         }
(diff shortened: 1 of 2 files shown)


wirew0rm (approved):
Looks good, using identity based equalness should be good here with model being final and initialized with a new Instance for every TableViewer.

RalphSteinhagen:
@dedeibel thanks for the PR! The identity-based comparison is also OK from my point of view. We are missing updated/custom `hashCode()`, `equals()`, and meaningful `toString()` functions here and there. Thus, these types of PR are much appreciated! :+1: 

N.B. as Obi-Wan once said: _"Trust_the clang-formatter"_ ... and the QA checks for that matter. :-)

Do you think you could upgrade the unit-test to improve the coverage also for this modified diff code? 

I am not only a fan of good unit-tests, but experience has shown that these make the code much more reliable as well as easier and more predictable to refactor or modify functionality. And as always, keep up the good work @dedeibel!

PR #348, opened 2021-01-28
Move from Travis to Github Actions

wirew0rm:


RalphSteinhagen (approved):
Very nice. I like the fact that the `pom.xml` became cleaner and the scripts more concise. Well done! :+1:

PR #349, opened 2021-01-28
Financial Positions and Orders added.

raven2cz:
Financial Positions and Orders full support added. 
The financial position renderer extension point added.
Full samples for positions to the historical daily charts and realtime trading simulation positions processing. Realtime calculation between positions and mapping to the renderer.
Basic execution platform and much more added to the samples for usage with chart-fx financial components.


RalphSteinhagen:
@raven2cz sorry for getting back late to you. 

I just started to review your PR. Overall it looks quite nice. :+1: :smile: 

There seems to be a lot of new financial-specific logic added. I presume this is/has to be all UI related. If you need a lot more of this type of code, maybe we could add these in another sub-module?


Quick early feedback on closing `FinancialRealtimeCandlestickSample`:
```Java
[JavaFX Application Thread] INFO de.gsi.chart.samples.financial.AbstractBasicFinancialApplication - requested demo to shut down
Exception in thread "Thread-5" java.lang.IllegalArgumentException: java.nio.channels.ClosedChannelException
	at de.gsi.chart.samples.financial.service.SimpleOhlcvReplayDataSet.lambda$getDataUpdateTask$1(SimpleOhlcvReplayDataSet.java:220)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.nio.channels.ClosedChannelException
	at java.base/sun.nio.ch.FileChannelImpl.ensureOpen(FileChannelImpl.java:150)
	at java.base/sun.nio.ch.FileChannelImpl.read(FileChannelImpl.java:209)
	at de.gsi.chart.samples.financial.service.SCIDByNio.loadOhlcvItemRealtime(SCIDByNio.java:194)
	at de.gsi.chart.samples.financial.service.SCIDByNio$1.get(SCIDByNio.java:167)
	at de.gsi.chart.samples.financial.service.SimpleOhlcvReplayDataSet.tick(SimpleOhlcvReplayDataSet.java:112)
	at de.gsi.chart.samples.financial.service.SimpleOhlcvReplayDataSet.lambda$getDataUpdateTask$1(SimpleOhlcvReplayDataSet.java:205)
	... 1 more
```

It seems to be just in the tests, so this messy clean-up on exit would be sort-of-OK. I wouldn't recommend these potentially dangling resource situations in a production environment. ;-)

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/execution/AbstractExecutionPlatform.java
```diff
@@ -0,0 +1,137 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: as above

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/execution/BacktestExecutionPlatform.java
```diff
@@ -0,0 +1,98 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: remove, please.

There seems to be a lot of financial non-UI domain functionality. Should one consider refactoring this into another dedicated sub-module, like 'dataset-financial'?

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/execution/BasicOrderExecutionService.java
```diff
@@ -0,0 +1,74 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: ...

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/order/PositionFinancialDataSet.java
```diff
@@ -0,0 +1,192 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: ...

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/order/ResolveOrderService.java
```diff
@@ -0,0 +1,168 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: ...

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/order/ResolvePositionService.java
```diff
@@ -0,0 +1,157 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: ...

chartfx-samples/src/main/java/de/gsi/chart/samples/financial/service/plan/MktOrderListTradePlan.java
```diff
@@ -0,0 +1,82 @@
+/**
+ * LGPL-3.0, 2020/21, GSI-CS-CO/Chart-fx, BTA HF OpenSource Java-FX Branch, Financial Charts
+ */
```
RalphSteinhagen, line 3: ...

RalphSteinhagen (dismissed):
Overall, this looks very good! :+1: 

Some minor comments:
  - if you could remove the additional license header as these are redundant.
  - I would merge this as (provided the first point is done), but would you consider the following: there seems to be a lot of usefull financial domain knowledge and related code in the PR which make some part look like a fully fledged application. Would it be conceivable to refactor the non-UI items into another new sub-module, for example, dataset-fincancial (package: de.gsi.dataset.financial.[..]? Presently, this is OK, but regarding long-term maintenance -- and I am on a limb that you might contribute more ;-) -- this will probably grow. I am OK with that but cannot assess how "big" this may become. Please let me know.

raven2cz:
> There seems to be a lot of new financial-specific logic added. I presume this is/has to be all UI related. If you need a lot more of this type of code, maybe we could add these in another sub-module?

**Add sub-module)**
I though about this around Christmas. There are two problems:
- Trading platforms are too complex and it cannot be part of the chart-fx library. For example we are using kotlin now and many domain objects are already written as data kt classes. 
- Chart-fx has to be prepared for more trading platform approaches and it has to be still graphical visualization library only. 

According to these problems. I tried to create very simple and demonstrated domain objects and services which are strictly required as basement for testing and demonstration of GRAPHICAL trading chart components (which belongs to the chart-fx).
I tried to involve all necessary services which will be necessary for ALL next planed trading components. The list of planned graphical components is following:
- orders manual manipulation lines
- volume profiles
- footprint profiles
- market profiles
- better subplot integration (open topic for you too)
- TimeIndicesAxis (still based on indices for X axis with sophisticated labeling with timestamps/aggregation of timegroups)
- better visualization of areas between minimal and maximal percentiles. And next small changes....
- it's all. No next components are planned.

I hope that services are almost finished for these necessary graph. components.
Maybe I have one idea, it is about discussion with you.
In my opinion, these services and domain objects are just for examples and "inspiration" have to use graphical financial components. So, it seems that it STILL belongs to the **samples**. But it needs separate package, because there is a lot of staff.
Maybe to create new SEPARATE package in the samples?

**src/main/java/de/gsi/financial/samples**

Add insert the components there. In future, we can extend these samples much more. 
FYI: These components are not necessary for graphical financial UI, it is still samples, but without it, it cannot be tested, presented, nothing. It is some example basement for correct functionality which is included...

RalphSteinhagen:
@raven2cz thanks for your detailed explanation. I understand and agree with your point.

I am not too worried about using Kotlin. We are already using packages like [Javalin](https://javalin.io/) and [OkHttp](https://github.com/square/okhttp) that rely on a mix of Java and Kotlin and that IMHO work well together. 

The thought was more -- based on my experience with earlier projects -- that sometimes new functionality is incrementally added one piece at a time while one tends to forget to -- or does not have the time for -- evaluating whether this could be incorporated in a possibly more generic way. You rightly pointed out, that many of the things are not finance specific and could be beneficial for other applications as well.
Our line-of-thought was/is how to generalise what could be also used elsewhere (w/o having to adopt a financial domain language) and how to modularise these and other (e.g. in our own case accelerator/particle physics) more specialised functionalities. For example, we have been evaluating for a long time whether to refactor the dataset module out of chart-fx, because this is also being used as a domain object in our communication middle-ware which we want to keep UI-free. 
I am sure we'll find a solution and I hope that you'll be open -- once we found a nice/strategic solution --  for refactoring this, if necessary.

For the time being, this isn't urgent. Provided you could remove the superfluous license headers (minor issue) and you confirm that you are also ok with the present state/PR, then I would merge this right away. Let me know.

W.r.t. the additional topics (project structure, Kotlin, Coverity issues etc.): we could discuss this in a separate thread at a later stage.

Thanks again for your PR and contributions. These are much appreciated!

chartfx-chart/src/test/java/de/gsi/chart/renderer/spi/financial/utils/CalendarUtils.java
```diff
@@ -114,4 +114,27 @@ public static Calendar createByDateTime(String datetimePattern) throws ParseExce
...
+     *
+     * @param datetimePattern String
+     * @return calendar interval instance
+     * @throws ParseException parsing fails
+     */
+    public static Calendar createByTradeStationDateTime(String datetimePattern) throws ParseException {
+        if (datetimePattern == null) {
+            throw new ParseException("The resource datetime pattern is null", -1);
+        }
+        SimpleDateFormat sdf = new SimpleDateFormat("MM/dd/yyyy HH:mm");
```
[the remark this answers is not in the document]
wirew0rm: I think with this format string there would be no problems, but adding a locale won't hurt.

The choice of format string seems to stem from the OhlvcItem representation, if this is an externally used format I think this would justify not using ISO format, otherwise I agree that it would be preferable. But in the end a question of preference.

Javafx properties are limited to the javafx thread so I don't know if it applies here.
For formatting (as opposed to parsing) i think introducing CSS styleables would be a good idea independently of this PR.

chartfx-chart/src/main/java/de/gsi/chart/renderer/spi/financial/PositionFinancialRendererPaintAfterEP.java
```diff
@@ -0,0 +1,220 @@
...
+    @Override
+    public void paintAfter(OhlcvRendererEpData d) {
+        if (d.index == d.minIndex) {
+            initByDatasetFxStyle();
+        }
+        long xcorr = Math.round(d.ohlcvItem.getTimeStamp().getTime() / 1000.0);
+        PositionRendered position = ((PositionRenderedAware) ds).getPositionByTime(xcorr);
+        if (position == null)
+            return;
+        d.gc.save();
```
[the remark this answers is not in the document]
raven2cz: It seems that whole Codev coverage was made before pushed junit tests. The EP class has 100% coverage. I try to change some line in the test if the coverage will be recalculated.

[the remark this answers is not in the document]
raven2cz: The whole class has 100% coverage, this is nonsense. Maybe the whole codecov was started before pushed junit tests. I will try to add some line to the test if the codecov will be activated and recalculated. In any case, whole class has 100%.

chartfx-chart/src/test/java/de/gsi/chart/renderer/spi/financial/utils/CalendarUtils.java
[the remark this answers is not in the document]
raven2cz: The format was removed. It was used just internally for FinancialTestUtils.

raven2cz (commented):
All points are done. But there is some deploy problem. I hope that it is some different problem, not with my modifs.
The old codev points will be correct in the next PR. 
The financial components for testing/samples were moved to the financial/samples package.

RalphSteinhagen (approved):
@raven2cz great and lgtm! :+1:

The issue with the deploy task is on our side and will be taken care of by us.

Thanks again for this great PR!

PR #362, opened 2021-02-16
added Codacy-based coverage diagnostics

RalphSteinhagen:


.github/workflows/ci.yml
```diff
@@ -89,10 +93,10 @@ jobs:
           path: ~/.m2
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
-      - name: Set up JDK11
-        uses: actions/setup-java@v1
+      - name: Set up the JDK
+        uses: actions/setup-java@v1.4.3
         with:
-          java-version: 11
+          java-version: ${{ matrix.java_reference }}
```
wirew0rm, line 99:
```suggestion
          java-version: 11
```

we cannot use the matrix variable here, because its scope is bound to the test job.
Maybe something like:
```suggestion
          java-version: ${{ jobs.test.matrix.java_reference }}
```
could also work but I havn't tested.
Or add an [output](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs) to the test job.

.github/workflows/ci.yml
```diff
@@ -101,9 +105,9 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GPG_PASS: ${{ secrets.GPG_PASSPHRASE }}
       - name: Set up JDK to publish to OSSRH
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v1.4.3
         with:
-          java-version: 11
+          java-version: ${{ matrix.java_reference }}
```
wirew0rm, line 110:
```suggestion
          java-version: 11
```
same as above.

wirew0rm (dismissed):
The requested changes should fix the publish step. The rest looks good. I like that the reference java version is now not an arbitrary number anymore :+1:

wirew0rm (approved):
perfect :+1:

PR #367, opened 2021-02-18
Fixes issue366

RalphSteinhagen:


ennerf (commented):
Without having done a deep analysis of the locking, the logic looks correct. I ran the test on a 12/24 cpu with various combinations and could not make it fail.

The synchronized blocks are however worrying, and the performance is quite a bit worse. I uploaded a video of comparing this PR (left) to 11.2.5 (right):

[chart-fx #367](https://www.youtube.com/watch?v=x-XzF9atdak)

The calculator shows how often the same plot is open on both versions. Each plot has 3x 6k points coming in at about ~640 Hz.

RalphSteinhagen:
moved stampedLock.read(..) into readLockGuard (avoids 'synchronized')
N.B. read(Un)Lock(..) still works but its use should be deprecated compared to the lock guards due to worse performance

qualitate improvement using DataSetLockStressTest running over 10 s:
<details><summary>click for details</summary>

before :
nWrites = 2320 nReads = 7322971
nWrites = 1940 nReads = 9190137
nWrites = 2120 nReads = 8164484
nWrites = 2320 nReads = 7523551
nWrites = 9280 nReads = 1869435
nWrites = 1980 nReads = 8683115
nWrites = 2280 nReads = 7463398
nWrites = 2020 nReads = 9642005
nWrites = 3080 nReads = 7642314
nWrites = 2140 nReads = 8356819

after:
nWrites = 5063940 nReads = 676150
nWrites = 5771120 nReads = 707951
nWrites = 5420000 nReads = 698665
nWrites = 5821140 nReads = 706828
nWrites = 5771180 nReads = 724168
nWrites = 4870600 nReads = 705715
nWrites = 6336320 nReads = 714426
nWrites = 5635140 nReads = 671819
nWrites = 5712100 nReads = 682183
nWrites = 4741540 nReads = 631645
</details>

Read-lock access (through the `readLockGuard(..)` should be as fast as the underlying `StampedLock`
The remaining differences are IMHO due to the test systematics and the remaining workaround for making the `StampedLock`write lock reentrant. Hope that I have found a better compromise between thread-safety and read-write lock performance. 

@ennerf if you could revalidate the new PR version agains your test scenario. Thanks in advance and much appreciated. :+1:

ennerf:
I just tried this version, but I immediately get into a deadlock. I can't even open a single chart for more than a few seconds. The `DataSetLockStressTest` still passes correctly though.

I'm using a custom data set for sorted time series data with the following add function:

```java
public void add(double x, double y) {
    lock().writeLockGuard(() -> {
        xValues.put(x);
        yValues.put(y);
        recomputeLimitsInternal();
    });
    fireInvalidated(addEvent);
}
```

RalphSteinhagen:
@ennerf thanks for your follow-up. I could reproduce your observation using the example 'de.gsi.chart.samples.legacy.ChartHighUpdateRateSample' which updates the DataSet @1kHz (N.B. in-build Chart-fx rate-limiter limits this to <50Hz).

replaced writeLock() internally with spinning StampedLock:tryWriteLock(..) the unit-tests and sample seem to work (for my part) but would be nice to see whether your use-case is still OK.


N.B. avoids dead-lock
de.gsi.chart.samples.legacy.ChartHighUpdateRateSample is a good test-case to reproduce the write-race-condition and prove of the dead-lock.

ennerf:
The performance is maybe slightly better than with the first fix, but it's tough to tell visually. Still noticably worse than pre-fix.

RalphSteinhagen:
@ennerf based on the stress tests the max write-lock frequency is now around 2-3 kHz...

In your example: how many writer/readers and update/read rates do you typically have?

Presently -- apart from getting the `Thread.currentThread()` -- it's pretty much bare StampedLock performance. The only (remaining) difference is that the readLock is now done locally and without `AtomicInteger` read-counter that are more efficient but cause the read lock/unlock race.... 

Have to think about whether one can still use atomics and only use the stampedLock.readLock() only for the first/critical reader...

ennerf:
1 writer + 1x UI thread reader w/ the concurrent rendering parts. The update rate for each plot is 1-4 x 0.1-1KHz.

RalphSteinhagen:
I'll give it another try with atomics-only ...

RalphSteinhagen:
modified read-lock and re-introduced atomic guards

only the first reader acquires a lock, subsequent readers increase atomic counter, and only the last remaining reader unlocks.

detects race-condition on initial lock and final unlock

ennerf (approved):

chartfx-dataset/src/main/java/de/gsi/dataset/locks/DefaultDataSetLock.java
```diff
@@ -36,24 +37,25 @@
...
+// Runnable used as functional interface
 public class DefaultDataSetLock<D extends DataSet> implements DataSetLock<D> {
     private static final long serialVersionUID = 1L;
     private final transient StampedLock stampedLock = new StampedLock();
-    private transient long lastReadStamp;
-    private transient long lastWriteStamp;
-    private transient Thread writeLockedByThread; // NOPMD
-    private final transient AtomicInteger readerCount = new AtomicInteger(0);
-    private final transient AtomicInteger writerCount = new AtomicInteger(0);
+    private final AtomicLong lastReadStamp = new AtomicLong(-1L);
```
ennerf, line 54: was there a reason why the previous fields were transient?

dedeibel:
I observed no more test failures with this version and could not see of any loopholes at the moment. Thanks.

PR #369, opened 2021-02-24
Remove Codacy CI steps

wirew0rm:
removes running the codacy analyzer and uploading the found defects to
github as sarif files. The update of the action broke the sarif format
and the information was redundant with the one provided by the codacy
bot.


RalphSteinhagen (approved):
OK. as discussed.

PR #375, opened 2021-03-05
ci: allow to restart deployment

wirew0rm:
When the deployment is restarted after having already partialy deployed,
the existing artifacts on github block the action from uploading the
artifacts to ossrh/mavencentral.
This commit allows the github action to fail, so that the maven central
artifacts are uploaded anyway. The downside is that failures to upload
to github will not be indicated correctly, but as maven central is our
main distribution channel, this should be ok.


RalphSteinhagen (approved):
OK

Issue #378, opened 2021-03-07
AutoRange axis in CircularDoubleErrorDataSet


RossHS:
Hello,

I am using a CircularDoubleErrorDataSet to display real-time data with a given 400 points buffer size and I would like to know if it even possible to implement autoRanging of the yAxis? What i mean. If you look at the first screenshot you may see the correct auto range for y axis. (For testing, I created a linear function growing in time with an initial point of (200;200) and auto range work correct). 
![RollingBufferSample](https://user-images.githubusercontent.com/30704653/110248153-b9856280-7f80-11eb-8703-8acead91cad6.jpg)
But over time, when the buffer overflows and we no longer see y-point 200, auto range locked at this minimum value and it doesn't work the way I want
![RollingBufferSample1](https://user-images.githubusercontent.com/30704653/110248306-6233c200-7f81-11eb-8078-498406d6d2e2.jpg)

Do you have any practice to solve this kind of problem?

Thanks for the great lib!

RossHS:
Little update. Version 11.2.3 does not have this problem, everything works as it should. Thus, this is a bug of version 11.2.4. As I understand in version 11.2.4, the values of the min and max in the CircularDoubleErrorDataSet::getAxisDescription are calculated incorrectly. 

screenshots of the same code, but with the library of the previous version:
![image](https://user-images.githubusercontent.com/30704653/110249957-e7bb7000-7f89-11eb-8705-b26a349c8c4f.png)
![RollingBufferSample3](https://user-images.githubusercontent.com/30704653/110249963-eab66080-7f89-11eb-81c6-bdcc97d11b36.jpg)

RalphSteinhagen:
@RossHS thanks for spotting and reporting. This may have slipped in with one of the last PRs. 

I will have a look at it and make a new release by the end of the day.

yezhengli-Mr9:
Dear @RossHS ,
How does your dataset update mechanism code look like? #647  For your use case of `CircularDoubleErrorDataSet`,
are you using 
https://github.com/fair-acc/chart-fx/blob/ed86318c5a1bde23d285633a94860648aee5fb32/chartfx-samples/src/main/java/io/fair_acc/sample/chart/PaddedAutoGrowAxisSample.java#L40-L62
or codes like (but does not feel that relevant to `CircularDoubleErrorDataSet`:
[...]

PR #379, opened 2021-03-08
fixes autoranging for CircularDoubleErrorDataSet (issue #378)

RalphSteinhagen:
fixes autoranging for CircularDoubleErrorDataSet (issue #378) 
+ additinal unit-tests and follow-up fixes


chartfx-dataset/src/main/java/de/gsi/dataset/spi/CircularDoubleErrorDataSet.java
```diff
@@ -206,6 +229,23 @@ public CircularDoubleErrorDataSet reset() {
 
     @Override
     public DataSet set(final DataSet other, final boolean copy) {
-        throw new UnsupportedOperationException("copy setting transposed data set is not implemented");
+        lock().writeLockGuard(() -> other.lock().writeLockGuard(() -> {
```
wirew0rm, line 232: Why is the write lock (as opposed to a read lock) on `other` necessary? From logic I would say that other is not changed, so other threads should be able to read simultaneously? The tests also pass if I change it to `other.lock().readLockGuard(...)` but I think it might be necessary to prevent deadlocking when the code inside the guard needs a write lock.  If it is necessary, maybe add a small comment with a short explanation, so it will not be carelessly removed in the future?
As this is consistent through the whole codebase, I think we can also take a look at this separately and leave this PR for the fix of the reported issue.

wirew0rm (approved):
Looks good. I like the restructured and generalized tests :+1:

dedeibel:
Also works with my codebase. Thanks.

Sorry for stirring this up…

chartfx-dataset/src/main/java/de/gsi/dataset/spi/CircularDoubleErrorDataSet.java
wirew0rm: Otherwise `dataset.set(dataset)` will deadlock because the read  lock cannot be upgraded to a write lock. This can also happen indirectly e.g. when using dataset wrappers.

PR #384, opened 2021-04-02
bump to restyled/restyler-clang-format:12.0.0

RalphSteinhagen:


raven2cz:
Hi Ralph, 
just one question to clang. Is it really correct?
IndentCaseLabels: false 
According to every development env or books etc. This is not valid visualization of switch / case structure. And always I have problems with this settings.

RalphSteinhagen:
@raven2cz  thanks for having a look at this PR. 

The original modification was by @wirew0rm in one of our other projects because we bumped our internal llvm use to a newer version. This PR is to synchronize the settings among our public GitHub projects.

According to the [clang-format doc](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) the `IndentCaseLabels: false` should be correct. At least our IDE and manual use of clang-format does not complain and does what it advertises i.e. aligns the `case xxx:` statements with the leading `switch` statement.

Regarding the value 'true' vs. 'false': this has been pre-determined by our internal controls group and/or synchronised with our C++ style (strongly inspired by Qt).

For me, the actual value isn't that important as long as it is consistent across our projects and people apply the same formatter. The latter is important because it limits the number of line changes one has to read just because of formatting changes. 

Since most of us do not program on their hand-held devices anymore and have >24" wide-format screens, I prefer to have rather longer than more lines (as long as they do not mix topics/concepts) because this helps to skip lines more easily that are not that important for the flow of the algorithm. But with modern IDEs the differences become also a bit more blurry ...

wirew0rm (approved):
:+1: 
It is a bit annoying that clang-format is not really great at backwards compatibility, but that seems to be a deliberate design choice on their end. Since it is the default version used by intellij, I think we can assume it as mainstream enough.

PR #388, opened 2021-04-13
Add issue templates

wirew0rm:
Adds an issue template, linking to the discussions feature and to gitter for non-bug issues.


RalphSteinhagen (approved):
looks good. Thanks for the improvement... we should use the issues primarily for bugs and discussion feature for Q&A and features... :+1:

Issue #409, opened 2021-05-02
Error using Zoomer Plugin - javafx.controls does not export com.sun.javafx.scene.control.behavior


nkoding:
**Describe the bug**
Hi, 

I've set up a JavaFX project and load the chart-fx dependency via maven.
In my chart class I'm using the Zoomer plugin like this:

`chart.getPlugins().add(new Zoomer());`

Without the zoomer plugin everything works fine. But with it I encounter the following error:

```
java.lang.IllegalAccessError: superclass access check failed: 

class impl.org.controlsfx.behavior.RangeSliderBehavior (in module org.controlsfx.controls)
 cannot access class com.sun.javafx.scene.control.behavior.BehaviorBase (in module javafx.controls) 
 because module javafx.controls 
 does not export com.sun.javafx.scene.control.behavior 
 to module org.controlsfx.controls
```
I've tried the VM-Args mentioned in Issue #6 and added this directive:
```
--add-opens
javafx.controls/com.sun.javafx.scene.control.behavior=ALL-UNNAMED
```

Also while trying stuff I found out that the modules javafx.fxml and javafx.web are producing those errors:
java.lang.module.FindException: Module javafx.web not found
java.lang.module.FindException: Module javafx.fxml not found

So I removed them and dont get the "not found"-errors anymore. 

Unfortunately the IllegalAccessError still persists.

Any help would be very appreciated.
Thanks in advance!

**Environment:**
 - OS: Ubuntu 20.04
 - Java version: 11
 - JavaFx version: 13
 - ChartFx version: 11.2.5

wirew0rm:
Hey,

the vm arguments are slightly different depending on which dependencies end up on the module path vs the classpath. If you read the error message, it mostly corresponds to the missing/incorrect vm argument.
In your case this should be `--add-exports=javafx.controls/com.sun.javafx.scene.control.behavior=org.controlsfx.controls`.
Since the option stems from our controlsfx dependency I will also refer you to their [wiki page about debugging jpms dependency problems](https://github.com/controlsfx/controlsfx/wiki/Using-ControlsFX-with-JDK-9-and-above#understanding-exceptions).

I'll try to update the README with more up-to-date parameters, but there will probably be corner cases where you have modify them. Thankfully, once you figured out how to read the exceptions, that is pretty straightforward.

I hope this helps,

Alex

wirew0rm:
[...]

PR #414, opened 2021-05-06
Fix data reduction for data containing NaN values

wirew0rm:
For charts with the ErrorDataSetRenderer with setNaNAllowed(true) and data reduction enabled, there the last data point before a gap caused by a NaN value was dropped (see #413 ).
Additionally when zooming out, there where spurious lines after the gap caused by the reduction algorithm not being correctly reset encountered by me and also independently reported by @dedeibel.

This PR adds a unit test for the DefaultDataReducer for all different error modes, fixes the bugs in the DefaultDataReducer and also adds the relevant parameters and sample data to the ErrorDataSetRendererStylingSample.

I compared performance with the ErrorDataSetRendererSample and the ErrorDataSetRendererStylingSample and could not make out a measurable difference.

Additionally I updated the jpms java-vm arguments in the README and linked to some troubleshooting resources there in response to #409.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ as jvm flags. Adding the following to the java command line call or your IDEs ru
 modules available and accessible to chartfx:
 
 ```
---add-modules=javafx.swing,javafx.graphics,javafx.fxml,javafx.media,javafx.web
+--add-modules=javafx.graphics,javafx.fxml,javafx.media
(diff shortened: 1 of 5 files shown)


RalphSteinhagen (approved):
Very nice and especially the new unit-test and ErrorDataSetRendererStylingSample extensions are very good additions.

PR #415, opened 2021-05-10
Financial Footprint Chart Renderer added.

raven2cz:
**Footprint Chart Information**
Footprint chart is a type of candlestick chart that provides additional information, such as trade volume and order flow, in addition to price. It is multi-dimensional in nature, and can provide an investor with more information for analysis, beyond just the security's price. This tool is a unique offering that is gaining popularity amongst leading charting software providers. 

Bid/Ask Footprint: Adds color to the real-time volume, for easier visualization of buyers and sellers probing the bid or ask. With this footprint, traders can see whether the buyers or the sellers are the responsible parties, for influencing a price move.

PR #417, opened 2021-05-12
Allow the ci to be run for external PRs

wirew0rm:
Uses the workflows defined on mater to run pull requests so that the
secrets are available.
https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/
For the coverity scan action we have to add an indirection via
workflow_dispatch -> workflow_run to have them run with the api token.
https://stackoverflow.com/a/65081720/2411625

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 ---
 name: Java CI
 
-on: [push, pull_request]
+on: [push, pull_request_target]
 env:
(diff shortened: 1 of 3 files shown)


RalphSteinhagen (approved):
Very good! Should also tackle the issue mentioned in PR #415

I presume we'll see whether this works as soon as we get another external PR from a forked user-repository

PR #418, opened 2021-05-14
fixes #383: inverted non-auto-range axes defaulting to +-1

### LINKED ISSUE #383: Fixed axis range does not work with inverted axis
Version: `chart-fx 11.2.5`

I'm trying to create a chart with a fixed range on an inverted axis. Creating a non-inverted one works

```Java
Chart chart = new XYChart();
Axis xAxis = chart.getFirstAxis(Orientation.HORIZONTAL);
xAxis.setAutoRanging(false);
xAxis.setAutoGrowRanging(false);
xAxis.set(0, 100);

Axis yAxis = chart.getFirstAxis(Orientation.VERTICAL);
yAxis.setAutoRanging(false);
yAxis.setAutoGrowRanging(false);
yAxis.set(0, 10);
```

image

but as soon as I add inversion, both X and Y axes are fixed to `[-1, +1]`. It does not matter whether it is only one axis or in which order the methods are called.

```Java
xAxis.invertAxis(true);
yAxis.invertAxis(true);
```

image

RalphSteinhagen:
N.B. axis range clearing doesn't seem to be necessary when inverting the axis
* added sample to illustrate the use of inverted axes (and also fix)
* added another (unrelated) sample to illustrate plotting of limits
* minor code clean-up (usage of 'var', unavoidable PMD/SONARLINT suppression).


wirew0rm (approved):
Looks good and I could find no regressions in the samples.

PR #428, opened 2021-06-12
fix: pic link in README.md

u386:
Fix in README.md the incorrect img ref to the figure ChartFx performance comparison for 2 Hz


RalphSteinhagen (approved):
@u386 well spotted and thanks for the PR!

PR #436, opened 2021-06-23
Use user agent css files and YValueWatchIndicator improvements

### LINKED ISSUE #427: WatchValueIndicator: css styling
**Describe the bug**
When styling the chart with JavaFX CSS, I can't get the `value-indicator-label`, `value-indicator-line`, `value-watch-indicator-label`, `value-watch-indicator-label` or their indexed variants to apply. If I subclass the respective `YWatchValueIndicator` and similar and assign IDs to line and label with `setID`, referencing these classes in the CSS does give the expected result with the CSS applied. Other classes like `chart-plot-background` are working as expected.

Also, the data value shown by `YWatchValueIndicator` appears _below_ the marker triangle, like so: 
misaligned value

With coloured `label`:
with label box

**To Reproduce**
The indicator is added through a UI interaction to `chart.getPlugins()`. The CSS file is added to the JavaFX scene as a stylesheet.

**Environment:**
 - OS: Windows
 - Java version: AdoptOpenJDK 15.0.2.7-hotspot
 - JavaFx version: OpenJFX 15.0.1
 - ChartFx version: `charfx-chart` 11.2.2 and 11.2.5 (I upgraded to check if it's a version thing)

wirew0rm:
#  Use user agent style sheet for default css

When adding the default css to the charts getStylesheets() list, it
effectively makes it impossible for users to overwrite values in
chart.css from their scene's stylesheet. By setting it as the chart's
user style sheet, the default chart.css has a lower priority than the
stylesheets added by the user.

# YWatchValueIndicator improvements

- Use formater of axis instead of fixed formater
- Calculate size of the Marker based on label size.

# Open Points:
- Testing with real applications and not just the samples
- <strike>Add a sample which provides its own css file </strike>
- The EditAxisPlugin and the GridRenderer (only for taking snapshots of the grid lines to cache dashed lines) still add stylesheets to the scenegraph, check if this can be avoided.

fixes #427

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,10 @@ jobs:
           key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-m2
       - name: Set up JDK ${{ env.JAVA_REFERENCE_VERSION }}
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v2.1.0
         with:
           java-version: ${{ matrix.java }}
+          distribution: 'adopt'
       - name: Test with Maven
         run: mvn -Dgpg.skip=true --no-transfer-progress --batch-mode -Drevision=${REVISION} -Dchangelist=${CHANGELIST} package
       - name: CodeCov coverage Reporter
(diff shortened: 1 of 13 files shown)


RalphSteinhagen (approved):
very good additions especially the CSS fixes/upgrades. lgtm! :+1:

PR #439, opened 2021-06-25
Fix YWatchValueIndicatorTest

wirew0rm:
Failed to adjust the expected string for the Indicator to the value returned by the tick label formatter in yesterday's PR, resulting in test failures.


RalphSteinhagen (approved):
LGTM!

PR #443, opened 2021-06-29
YWatchValueIndicator: prevent occlusion of value labels

wirew0rm:
By default, if there are multiple YWatchValueIndicators close to each other, their label texts occlude each other. This PR implements an optional flag to move the value labels on a per axis basis.

Axis on the left shows the old and now default behaviour, the Axis on the right showcases the new feature:
2021-06-29 15-23_marker_occlusion


RalphSteinhagen (approved):
Very nice addition! :+1:

@raven2cz this new feature might perhaps be also of interest for your use-case?

PR #446, opened 2021-07-06
fix DataSetMath integration and ErrorDataSetRenderer bug

RalphSteinhagen:
* integration: caused by `DIM_X` vs. `DIM_Y` mixup 
* graphing bug: cause by mid-range contour points not swapping their y coordinates


RalphSteinhagen:
Joint PR for fixing:
* integration: caused by `DIM_X` vs. `DIM_Y` mixup 
* fix DefaultDataReducer
  This was originally fixed in #271 but it turned out that the fix was
incomplete and only applied for data with xy errors. This fixes the same
issue for y error only datasets and the x data of xy error datasets.
* add test dataset for different error types to ErrorDataSetRendererSample
* ValueIndicator: disable dragging for non editable indicators
  solves issue reported in #448

The first commit reviewed by @wirew0rm, the second was jointly fixed, and the rest reviewed by @RalphSteinhagen 

merging with admin rights

PR #450, opened 2021-07-12
added peak-width-via-integral estimator

RalphSteinhagen:
Adds statistical width estimator inspired by Gaussian distribution `n`-sigma property/definition, i.e. for gaussian distributions:
* 1 * sigma -> 68.3% probability of individual sample point being within the distribution 
* 2 * sigma -> 95.5% probability [..],
* etc.

See `PeakWidthSample` example for details:
image


wirew0rm (approved):
Looks good :+1:

PR #456, opened 2021-07-12
non-linear time-axis proof-of-concept and DataSetMath additions with sample

RalphSteinhagen:
We have some cases where we need both a short- and long-term history. 
Having two separate plots would take too much screen real-estate. Here's a proof-of-concept for comment:

image

For the time being, there are only two ranges but the labels being updated according to the given sub-range. Production use requires some more additional unit-tests and ']0,1[' check/limiting.
N.B. the same principle could be used for 'magnifying glass'-type interactor...

Food for thought and open for comments... :smiley:


wirew0rm (approved):
Nice addition to the samples, which can maybe be moved to a proper axis implementation in the future. For now it can illustrate how to customize axis behavior from user code.

Also contains new DataSetMath functionality, which estimates the width of a peak by integrating the area under the peak until a certain threshold of the total integral is reached and a sample application for it:
image

PR #466, opened 2021-09-23
Fix padded auto grow axis

### LINKED ISSUE #465: AutoRangePadding w/ AutoRangeGrowing compounds every iteration
**Describe the bug**
DefaultNumericAxis: AutoRangePadding does not work with AutoRangeGrowing

I'd expect the below code to start at [0,10] and over time grow to [0, max*1.05]. What happens is that it adds 5% at every iteration and compounds to [0,>1E24]. 

image

The range is further increased whenever the window is resized.

image

**To Reproduce**
 
``` java
public class AxisRangeCheck extends Application {

    public static void main(String[] args) {
        launch(args);
    }

    @Override
    public void start(Stage primaryStage) throws Exception {

        var xAxis = new DefaultNumericAxis();
        var yAxis = new DefaultNumericAxis();
        var chart = new XYChart(xAxis, yAxis);

        yAxis.setAutoRangePadding(0.05);
        yAxis.setAutoRanging(false);
        yAxis.setAutoGrowRanging(true);
        yAxis.set(0, 10);

        var ds = new DoubleDataSet("", 1000);
        chart.getDatasets().addAll(ds);
        new Thread(() -> {
            for (int i = 0; i < 1000; i++) {
                ds.add(i, i);
                try {
                    TimeUnit.MILLISECONDS.sleep(20);
                } catch (InterruptedException e) {
                }
            }
        }).start();

        Scene scene = new Scene(chart, 800, 600);
        primaryStage.setScene(scene);
        primaryStage.show();

    }

}
```

**Environment:**
 - OS: Windows 10
 - Java version: Zulu 16
 - JavaFx version: 16
 - ChartFx version: 11.2.5

wirew0rm:
With padding enabled auto grow ranging was broken (see #465, thanks to
@ennerf for reporting) because the padding was applied to the auto
range stored in the axis. Because of this the padding was added each
time the axis range was updated with grow ranging enabled.

This commit changes Abstract axis to directly use the result of
`autoRange()` instead of overwriting the autoRange property with it.

In XYChart the autoGrow code had to be changed to use the previous
computed range instead of the displayed range (which would include the
padding, leading to pileup).

fixes #465

Also adds PaddedAutoGrowAxisSample based on the sample from the bugreport extended by adding a few plugins and a little bit more interesting data to check a few additional cornercases.

Note that the padding at zero is not added, which is intended behaviour; padding never pushes the bounds below/above zero.

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
@@ -448,8 +448,8 @@ public class XYChart extends Chart {
             return;
         }
         final boolean oldAutoState = axis.autoNotification().getAndSet(false);
-        final double oldMin = axis.getMin();
-        final double oldMax = axis.getMax();
+        final double oldMin = axis.getAutoRange().getMin();
+        final double oldMax = axis.getAutoRange().getMax();
         final double oldLength = axis.getLength();
 
         final boolean isHorizontal = axis.getSide().isHorizontal();
@@ -482,8 +482,8 @@ public class XYChart extends Chart {
         }
 
         if (axis.isAutoGrowRanging()) {
-            axis.getAutoRange().add(axis.getMin());
-            axis.getAutoRange().add(axis.getMax());
+            axis.getAutoRange().add(oldMin);
+            axis.getAutoRange().add(oldMax);
         }
 
         axis.getAutoRange().setAxisLength(axis.getLength() == 0 ? 1 : axis.getLength(), side);
--- a/chartfx-chart/src/main/java/de/gsi/chart/axes/spi/AbstractAxis.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/axes/spi/AbstractAxis.java
@@ -338,7 +338,7 @@ public abstract class AbstractAxis extends AbstractAxisParameter implements Axis
     }
 
     /**
-     * Called when data has changed and the range may not be valid any more. This is only called by the chart if
+     * Called when data has changed and the range may not be valid anymore. This is only called by the chart if
      * isAutoRanging() returns true. If we are auto ranging it will cause layout to be requested and auto ranging to
      * happen on next layout pass.
      *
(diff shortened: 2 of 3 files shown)


ennerf:
Thank you for the quick turnaround. This fixes the AutoRangePadding + AutoGrowRanging issue, but it still has some unexpected behavior. Changing the range to something more obvious, e.g., `yAxis.set(-30, +30)`


- does not enforce a minimum range of [-30,+30]
image

- does not snap back to the defined range once the peaks disappear from the data
image

RalphSteinhagen:
This is sort of the design behaviour:  the auto-grow feature presently work as being the joined range of the combined actual 'auto-range' of all DataSets associated with the axis and the axis previous range. I.e. it always grows and never shrinks. 

There have been some ideas floating around that after some time the ranges would automatically "shrink" or at least reduce to some smaller value that is closer to the instantaneous auto-range. 

Unfortunately, I haven't been able to sort out the required logic that works for most/all the corner cases:  e.g. how to deal with outliers/spikes and this has to be probably `Timer` based -- the question there is: what are suitable time scales when to start shrinking, how fast, etc. :thinking:

Ideas, suggestions, or implementations are welcome...

wirew0rm:
> Thank you for the quick turnaround. This fixes the AutoRangePadding + AutoGrowRanging issue, but it still has some unexpected behavior. Changing the range to something more obvious, e.g., `yAxis.set(-30, +30)`
>   * does not enforce a minimum range of [-30,+30]

You could use `yAxis.getAutoRange().set(-30,30)` for this effect in combination with autoGrowRanging. Probably we could add the user range to the range calculation, to make your more intuitive approach also work, but I'll have to check for side effects.

>   * does not snap back to the defined range once the peaks disappear from the data

If i understand you correctly, this would be the behaviour expected from the regular auto ranging, minus the minimum range functionality. As an ugly hack, the minimum range could be enforced there by having an empty dataset with getAxisDescriptions() overridden to always return [-30,30]. 

I think Ralph already described the intended behaviour of auto-grow and its limitaions, I could add, that you can use `yAxis.getAutoRange().clear()` (or .set(min,max)) to reset the chart to the current range (or your minimum range).

ennerf:
Thank you for the clarification. For now I've solved it by overriding `computeRange(min, max, axisLength, labelSize)`, but I assume this is a fairly common use case that may warrant api support.

As a non-experienced user I think I'd expect the following cases:

| case | set(min,max) | AutoRange  | AutoGrowRanging | Effect                                                             |
|:----------:|:---------------:|:------------:|:-----------------:|--------------------------------------------------------------------|
| 1        | X | O | O | fixed range that never changes                                     |
| 2        | O | X | O | automatically determines range for the currently available data    |
| 3        | O | any | X | same as 2, but does not retrace from largest value ever seen  |
| 4        | X |  X | O | fixed range that expands if the currently displayed data is out of bounds |
| 5        | X | any | X | same as 4, but does not retrace from largest value ever seen                       |

- AutoRange looks the same as AutoGrowRanging without the extra state, so AutoRange would already be implied?

ennerf:
More on-topic: The PR works and matches the functionality you described

ennerf (approved):

PR #481, opened 2022-01-15
add visibility state and toggling of datasets

ennerf:
This is an outline for what the toggle feature could look like. It's not ready for merging. Below are some questions I've run into so far:

1. What's the best way to omitt the hidden datasets in the axis range computation? Ideally something that could work on the Chart baseclass?
 
2. Is there a good way to deal with access to the dataset, e.g., `EditDataSet` without supporting it in every plugin?

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
@@ -258,9 +258,7 @@ public class XYChart extends Chart {
         ObservableList<DataSet> dataSets = this.getAllDatasets();
         // check that all registered data sets have proper ranges defined
         dataSets.parallelStream()
-                .forEach(dataset -> dataset.getAxisDescriptions().parallelStream()
-                        .filter(axisD -> !axisD.isDefined())
-                        .forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
+                .forEach(dataset -> dataset.getAxisDescriptions().parallelStream().filter(axisD -> !axisD.isDefined()).forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
 
         final ArrayDeque<DataSet> lockQueue = new ArrayDeque<>(dataSets);
         recursiveLockGuard(lockQueue, () -> getAxes().forEach(chartAxis -> {
--- a/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
@@ -258,7 +258,6 @@ public class XYChart extends Chart {
         ObservableList<DataSet> dataSets = this.getAllDatasets();
         // check that all registered data sets have proper ranges defined
         dataSets.parallelStream()
-                .filter(DataSet::isVisible) // TODO (ennerf): this does not get the dataset out of the axis. Is there a general way to do this w/o modifying the dataset itself ?
                 .forEach(dataset -> dataset.getAxisDescriptions().parallelStream()
                         .filter(axisD -> !axisD.isDefined())
                         .forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
@@ -458,7 +457,7 @@ public class XYChart extends Chart {
         final boolean isHorizontal = axis.getSide().isHorizontal();
         final Side side = axis.getSide();
         axis.getAutoRange().clear();
-        dataSets.forEach(dataset -> dataset.lock().readLockGuard(() -> {
+        dataSets.stream().filter(DataSet::isVisible).forEach(dataset -> dataset.lock().readLockGuard(() -> {
             if (dataset.getDimension() > 2 && (side == Side.RIGHT || side == Side.TOP)) {
                 if (!dataset.getAxisDescription(DataSet.DIM_Z).isDefined()) {
                     dataset.recomputeLimits(DataSet.DIM_Z);
--- a/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
@@ -257,8 +257,11 @@ public class XYChart extends Chart {
         // lock datasets to prevent writes while updating the axes
         ObservableList<DataSet> dataSets = this.getAllDatasets();
         // check that all registered data sets have proper ranges defined
-        dataSets.parallelStream().forEach(dataset -> dataset.getAxisDescriptions().parallelStream().filter(axisD -> !axisD.isDefined()) //
-                                                             .forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
+        dataSets.parallelStream()
+                .filter(DataSet::isVisible) // TODO (ennerf): this does not get the dataset out of the axis. Is there a general way to do this w/o modifying the dataset itself ?
+                .forEach(dataset -> dataset.getAxisDescriptions().parallelStream()
+                        .filter(axisD -> !axisD.isDefined())
+                        .forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
 
         final ArrayDeque<DataSet> lockQueue = new ArrayDeque<>(dataSets);
         recursiveLockGuard(lockQueue, () -> getAxes().forEach(chartAxis -> {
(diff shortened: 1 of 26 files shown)


chartfx-chart/src/main/java/de/gsi/chart/XYChart.java
```diff
@@ -257,8 +257,11 @@ public void updateAxisRange() {
         // lock datasets to prevent writes while updating the axes
         ObservableList<DataSet> dataSets = this.getAllDatasets();
         // check that all registered data sets have proper ranges defined
-        dataSets.parallelStream().forEach(dataset -> dataset.getAxisDescriptions().parallelStream().filter(axisD -> !axisD.isDefined()) //
-                                                             .forEach(axisDescription -> dataset.lock().writeLockGuard(() -> dataset.recomputeLimits(axisDescription.getDimIndex()))));
+        dataSets.parallelStream()
+                .filter(DataSet::isVisible) // TODO (ennerf): this does not get the dataset out of the axis. Is there a general way to do this w/o modifying the dataset itself ?
```
RalphSteinhagen, line 261: I think a good place to add the `.filter(DataSet::isVisible)` statement would be in `XYChart`
```
protected static void updateNumericAxis(final Axis axis, final List<DataSet> dataSets) {/*...*/}
```
This is the place where the chart loops through all DataSets that are relevant for a given axis and updates the auto-range and auto-grow-ranging.

chartfx-chart/src/main/java/de/gsi/chart/legend/spi/DefaultLegend.java
```diff
@@ -106,9 +107,17 @@ protected double computePrefWidth(final double forHeight) {
 
     public LegendItem getNewLegendItem(final Renderer renderer, final DataSet series, final int seriesIndex) {
         final Canvas symbol = renderer.drawLegendSymbol(series, seriesIndex, SYMBOL_WIDTH, SYMBOL_HEIGHT);
-        return new LegendItem(series.getName(), symbol);
+        var item = new LegendItem(series.getName(), symbol);
+        if(renderer.supportsVisibility()){
```
RalphSteinhagen, line 111: I wonder whether we want this optional or as a default requirement. This is a nice feature and not too much to ask for renderer implementers to include. 

Alternatively -- if implementers forget to add/implement this -- the data set wouldn't be taken into the auto-range calculation but still been drawn. Als an interesting effect but not sure whether this is really bad or actually good....

chartfx-chart/src/main/java/de/gsi/chart/renderer/Renderer.java
```diff
@@ -67,4 +67,14 @@
      * @return true (default) if data sets are supposed to be drawn
      */
     BooleanProperty showInLegendProperty();
+
+    /**
+     * Indicates whether the renderer supports dataset visibility.
+     *
+     * @return false (default) if the renderer does not support visibility.
+     */
+    default boolean supportsVisibility() {
```
RalphSteinhagen, line 76: as commented above... possibly this could be removed and made a default requirements for all renderer

chartfx-chart/src/main/java/de/gsi/chart/renderer/spi/ErrorDataSetRenderer.java
```diff
@@ -169,6 +174,11 @@ public Marker getMarker() {
             stopStamp = ProcessingProfiler.getTimeStamp();
             final DataSet dataSet = localDataSetList.get(dataSetIndex);
 
+            if(!dataSet.isVisible()) {
+                localDataSetList.remove(dataSetIndex); // TODO (ennerf): is this correct and/or necessary?
```
RalphSteinhagen, line 178: Since the default colour and styles are coupled to the data set index in the list, I'd just add this as a continue statement in the main loop, e.g.

```java
for (int dataSetIndex = localDataSetList.size() - 1; dataSetIndex >= 0; dataSetIndex--) {
    // [..]
    if(!dataSet.isVisible()) {
        continue;
    }
    // [..]
}
```
This way muted DataSets wouldn't change the colour and/or other styles of the other non-muted DataSets.

RalphSteinhagen:
Very nice PR and good addition to chart-fx! :+1:

chartfx-chart/src/main/java/de/gsi/chart/legend/spi/DefaultLegend.java
wirew0rm: If it is only possible to toggle to non-visible if the renderer supports it, I think it should be safe to do it this way. So i guess it really boils down to whether supporting this extra logic is more effort than just implementing it for all renderers.

ennerf: You're probably right. I made it a default requirement and added it to all renderers I could find.

chartfx-chart/src/main/java/de/gsi/chart/renderer/spi/ErrorDataSetRenderer.java
ennerf: Done. I got rid of some of the `drawnDataSets = new ArrayList(localDataSetList)` copies where every dataSet gets added anyways.

ennerf:
I added the suggested changes and ran through all samples. The IMO noteworthy ones are:

**EditDataSetSample**
Hiding a dataset does not remove the point selection. Without modifications it is possible to select both datasets, hide one, and then apply operations on hidden points resulting in indexing errors. I changed it such that points remain selected, but that any modifying operations are ignored.

**MetaDataRendererSample**
hiding random walk also hides the gaussy dataset. Maybe this is actually be the desired behavior?

chartfx-dataset/src/main/java/de/gsi/dataset/spi/AbstractDataSet.java
```diff
@@ -378,6 +377,11 @@ protected boolean equalValues(final DataSet other, final double epsilon) {
         return true;
     }
 
+    @Override
```
RalphSteinhagen, line 380: JavaFX dependence in non-UI module. The solution could e.g. look something like

```java
   // in the header
   private boolean isVisible = true;
   // [..]
    public boolean isVisible() { return isVisible; };

    public void setVisible(boolean visible) {
        if (visible != isVisible) {
            isVisible = visible;
            fireInvalidated(new UpdatedMetaDataEvent(this, "removed style"));
        }
    }
```
What do you think?

chartfx-chart/src/main/java/de/gsi/chart/legend/spi/DefaultLegend.java
```diff
@@ -106,9 +107,17 @@ protected double computePrefWidth(final double forHeight) {
...
+            item.setOnMouseClicked(event-> series.setVisible(!series.isVisible()));
+            item.pseudoClassStateChanged(disabledClass, !series.isVisible());
+            series.visibleProperty().addListener((obs, old, newValue) -> item.pseudoClassStateChanged(disabledClass, !newValue));
```
RalphSteinhagen: This would probably need to changeed to the non-UI event listener, something like:
```java
        series.addListener(evt -> {
            // possible with a guard to listen only to meta event changes and when the state actually changed
            item.pseudoClassStateChanged(disabledClass, !series.isVisible());        
        });
```

RalphSteinhagen (changes requested):
Hi @ennerf looks good.

It doesn't compile in the CI/CD due to JavaFX dependencies in non-UI modules (new BooleanProperty used for visibility state) but that IMO could be easily replaced by a simple `boolean isVisible` state and using the existing event notification mechanisms, e.g. in AbstractDataSet

```Java
    public boolean isVisible() { return isVisible; };

    public void setVisible(boolean visible) {
        if (visible != isVisible) {
            isVisible = visible;
            fireInvalidated(new UpdatedMetaDataEvent(this, "changed visibility"));
        }
    }
```
With these changes it should compile/work as expected. Please have a look (e.g. merge) also at the code-formatter (Restyled.io bot PR) and static-code analysis checks. 

After that this could be merged. Thanks for this very nice PR! :+1:

wirew0rm:
Sorry, I kind of missed the updates in this PR.

> **MetaDataRendererSample** hiding random walk also hides the gaussy dataset. Maybe this is actually be the desired behavior?

This is actually a bug in the sample I also discovered when I investigated the 11.2.6 axis regression. The gaussy dataset implements incorrect range information by overwriting AbstractTestFunction's `get()` function but not updating the range. So when you toggle visibiltiy of the other datasets, the axis range changes to something outside of the gauss functions x-value range, you probably can get it to show up again by manually zooming. I have a partial fix for that here, so don't worry about this one.

wirew0rm (approved):
Very nice feature, I really like how it turned out!

RalphSteinhagen (approved):

milo-gsi:
Thanks for providing that very helpful feature, Florian!
Just stumbled over it "accidentally" and I love it.

Issue #476, opened 2022-01-11
Value indicator triggers re-rendering of datasets


ennerf:
**Describe the bug**
Moving a value indicator triggers a layout of the entire chart. This is a big performance hit for large datasets.

**To Reproduce**
Run `ChartIndicatorSample` and move the indicator
 
**More Information**
The problem seems to be that the marker gets added to the plot foreground (see [AbstractSingleValueIndicator](https://github.com/GSI-CS-CO/chart-fx/blob/cd202c7c9a0b40f1e66b6d0fe8fd93fffbec5435/chartfx-chart/src/main/java/de/gsi/chart/plugins/AbstractSingleValueIndicator.java#L239-L253)), which triggers a layout of the chart area.

Removing the marker makes it work as expected.

```java
var indicator = new XValueIndicator(chart.getXAxis(), 0, "") {
    @Override
    protected void layoutMarker(double startX, double startY, double endX, double endY) {
        // don't add a marker to avoid re-rendering the chart area
    }
};
chart.getPlugins().add(indicator);
```

I'm not familiar enough with the scene graph layout. Is it possible to have it in the axis area without causing re-rendering? If not, do you think it would make sense to have a flag for hiding the marker for the few cases where it matters?

wirew0rm:
Hey, thanks for bringing this up, this was changed in 11.2.0 [here](https://github.com/GSI-CS-CO/chart-fx/commit/9f4a563ca77a09989668194ea7ce508930a0842c#diff-b981098b2e3524f8edd4c170b07cafcf85645ad8ad2d4112f025e14195c2a401L237-R250) because it was necessary to see to which axis an indicator belonged to when there were multiple axes. I guess there should be a way to make the node not trigger the relayout, but i'll have to investigate how to do that.

wirew0rm:
solved in #485

PR #485, opened 2022-02-08
Prevent relayouting of the chart on indicator change

wirew0rm:
Set the layout property of the indicator triangle to unmanaged s.t.
changes on the triangle will not invalidate the layout of its parent foreground pane.

Since the position of the triangle is already computed manually and
guaranteed to stay inside of the plot bounds, this should be the correct
solution here.

From the openjfx documentation for the `managed` property:
> Defines whether or not this node's layout will be managed by it's parent.
> If the node is managed, it's parent will factor the node's geometry
> into its own preferred size and layoutBounds calculations and will lay it
> out during the scene's layout pass. If a managed node's layoutBounds changes,
> it will automatically trigger relayout up the scene-graph to the nearest
> layout root (which is typically the scene's root node).
>
> If the node is unmanaged, its parent will ignore the child in both preferred
> size computations and layout. Changes in layoutBounds will not trigger
> relayout above it. If an unmanaged node is of type Parent, it will act as a
> "layout root", meaning that calls to Parent.requestLayout() beneath it will
> cause only the branch rooted by the node to be relayed out, thereby isolating
> layout changes to that root and below. It's the application's responsibility
> to set the size and position of an unmanaged node.
>
> By default all nodes are managed.
https://openjfx.io/javadoc/17/javafx.graphics/javafx/scene/Node.html#managedProperty

solves #476

@ennerf does this solve your indicator performance issue?

This might also apply to other nodes in the chart, especially with all the plugins that work on the overlay so we should check if there are more nodes to set to unmanaged.

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/plugins/AbstractRangeValueIndicator.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/plugins/AbstractRangeValueIndicator.java
@@ -72,6 +72,7 @@ public abstract class AbstractRangeValueIndicator extends AbstractValueIndicator
         setUpperBound(upperBound);
 
         rectangle.setMouseTransparent(true);
+        rectangle.setManaged(false);
         getChartChildren().addAll(rectangle, label);
     }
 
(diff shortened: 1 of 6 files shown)


ennerf:
That fixed it. Good find 👍

ennerf (dismissed):

ennerf (approved):

PR #492, opened 2022-03-07
Fix empty axis units so "[]" is no longer displayed

edkennard:
If no axis units have been specified there's no need to render empty square brackets "[]" after the axis name.


wirew0rm:
Hey, thank you for the PR.

This is actually the intended behaviour, as all axes should have units. If you do not want to display units on the axes (maybe because you display the unit somewhere else like in the legend or title) you can disable the display of the unit by setting it to `null`. There should probably be some documentation added for this.

If you never want units you can of course always override this functionality in your own axis implementation.

This is obviously an opinionated choice and we hope you can understand our point of view.

edkennard:
Hi @wirew0rm , I have revisited my code and found that the incorrect behaviour I was seeing is no longer there. This is probably because since yesterday I'm now deriving my own Axis classes, where before I was directly instantiating the DefaultNumericAxis class. I'm sure there was an issue doing the latter, but you are right, sorry to waste some of your time

wirew0rm:
I was just rechecking our samples and the `yAxis.setUnit(null)` worked for me (using `DefaultNumericAxis`), but I wouldn't have ruled out the possibility that some unrelated change might have broken this behaviour, so no worries, feedback from users is always appreciated, and even more if they show that they took some time to investigate and are willing to contribute back like you did.

Glad it works for you now, even though it should also work without subclassing the Axis.

PR #493, opened 2022-03-07
Fix style parser font behaviour

edkennard:
- In several places the default font was incorrectly named "Helvetia" instead of "Helvetica"
- The way the font name was being read by the StyleParser meant a blank "font=" wasn't handled correctly. The related tests have been improved accordingly

### DIFF
--- a/chartfx-chart/src/main/java/de/gsi/chart/renderer/spi/AbstractMetaDataRendererParameter.java
+++ b/chartfx-chart/src/main/java/de/gsi/chart/renderer/spi/AbstractMetaDataRendererParameter.java
@@ -16,7 +16,7 @@ import de.gsi.chart.renderer.spi.utils.ChartIconFactory;
 
 public abstract class AbstractMetaDataRendererParameter<R extends AbstractMetaDataRendererParameter<R>> {
     protected static final String STYLE_CLASS_LABELLED_MARKER = "chart-meta-data";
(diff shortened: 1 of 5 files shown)


RalphSteinhagen:
@edkennard thanks for this PR! :+1:

Yes, the intended use was 'Helvetica'... the other one is cursive...

RalphSteinhagen (approved):
Thanks! LGTM :+1:

edkennard:
@RalphSteinhagen Thanks for getting this merged - when will the next version be published to Maven Central?

RalphSteinhagen:
@edkennard we plan to make a micro-release this, latest next week. 

After that we will start with a new minor version since we have quite a few internal improvements pending (changes to the event handling structure -> moving to using the Disruptor, ...). This will also come with a GH organisation and authority change from `de.gsi' -> 'io.fair-acc'. This will help simplify our deployment process internal to GSI/FAIR.

Most of the rest will stay the same though. Enjoy!

P.S. If you like, maybe you could share your use-case in the discussion ... we always love feedback on how our little lib is being used.

edkennard:
@RalphSteinhagen ok great, thanks, the pending release will also include the PR which allows toggling of datasets, which we like.

I would be happy to add our use-case to the discussions area, will do that at a suitable moment in the next couple of weeks, once my current chart-related work is completed and merged into our product. My hope is to contribute further to the project too, as soon as time permits.

Cheers, Ed

Issue #508, opened 2022-04-14
Zoom plugin does not appear to correctly scale the y-axis for contour datasets


ajackson2681:
**Describe the bug**
I noticed while I was implementing my own project, that when I add the zoomer plugin to an XYPlot, that zooming in does not correctly scale the data for a contour dataset. This is easily reproducible by just running the Histogram2DimSample included in this repository. As you can see in the gif attached, when the chart zooms in, the dataset in the y direction appears to compress, even though the range on they y-axis is getting smaller. The x-axis does not appear to have this problem.

**Environment:**
 - OS: Pop!_OS 21.10
 - Java version: 11.0.14
 - JavaFx version: 13
 - ChartFx version: 11.2.6
 
![hist2d](https://user-images.githubusercontent.com/48104097/163447755-2a150353-8707-4954-bfdc-6c37bf6ea16c.gif)

wirew0rm:
fixed in #534

PR #519, opened 2022-05-17
Use DIM_Y for case DIM_Y in Histogram2 getGrid (#508)

ajackson2681:
Histogram2's implementation of getGrid was using DIM_X for they y projection as well as the x projection. This changes it to properly use DIM_Y.


RalphSteinhagen (approved):
@ajackson2681 well spotted and thanks for the quick-fix PR! Much appreciated. :+1:

PR #534, opened 2022-07-15
Package rename and bugfixes

wirew0rm:
Renames all packages and maven modules into the `io.fair_acc` package.
This will mean that all users of master/main branch will have to adjust the package names in their code.
This will be part of the upcoming 11.3.0 release.

Fixes some of the outstanding issues, see individual commit messages for explanations.


pacher:
I don't know about your versioning scheme, but maybe such a break everything release deserves a major version bump?
I understand that fixing imports is a trivial migration, but it is a migration nonetheless.

wirew0rm:
@pacher Thanks for getting involved. Basically in our versioning scheme, the second number is our 'major' version as the first one is only used to reflect the compatible java version, to allow us to maintain multiple versions for different java versions. At the moment there is only JDK11 supported, but we had to maintain the JDK8 version for a long time since modularization caused big code and packaging changes while many users could not simply update to JDK11. After the 11.3.0 release we we want to add a JDK17 branch (and 17.3.0 release), which will hopefully not diverge as much as the JDK8 branch did, but allow us to evaluate migration to a more modern language standard while still supporting users on java11.

Due to the changes in layout and event listener re-implementation it will probably be a bit more than just fixing imports. We try to make this as smooth as possible for our users, but we have accumulated an amount of workarounds and open issues with the current architecture that is becoming increasingly hard to maintain.

RalphSteinhagen (approved):
LGTM! Primarily tested on local build environment. Deployment to maven central to be checked once merged to main. :+1:

PR #536, opened 2022-07-25
mavencentral: update repo hostname

wirew0rm:
Follow-up to #534
As of feb '21, new mavencentral projects are located on s01.oss.sonatype.org instead of oss.sonatype.org.
With this PR it should be possible to deploy to mavencentral again.


wirew0rm:
the credentials also changed for the new system.

PR #160, opened 2020-04-19
Modularization

wirew0rm:
Efforts to make chartfx usable from jigsaw modularized projects as proposed in #156.

Before merging this the following points will have to be addressed:

## Jigsaw Compatibility

To make chartfx usable for modularized projects, there are some points which have to be considered according to this [blogpost](http://branchandbound.net/blog/java/2017/12/automatic-module-name/).

> Addressing these concerns is a matter of good hygiene. If you encounter one of these issues and you can't address those, don't add the Automatic-Module-Name entry yet. For now, your library is better off on the classpath. It will only raise false expectations if you do add the entry.

So before meging this branch, these issues should probably be adressed.

## internal JDK types

> Make sure your library doesn't use internal types from the JDK (run jdeps --jdk-internals mylibrary.jar to find offending code). JDeps (as bundled with Java 9 and later) will offer publicly supported alternatives for any use of encapsulated JDK APIs. When your library runs from the classpath on Java 9 and later, you can still get away with this. Not so if your library lives on the module path as automatic module.

```
jdeps --jdk-internals chartfx-chart/target/chartfx-chart-11.1.0-SNAPSHOT.jar 
chartfx-chart-11.1.0-SNAPSHOT.jar -> JDK removed internal API
   de.gsi.chart.utils.SimplePerformanceMeter          -> com.sun.javafx.perf.PerformanceTracker             JDK internal API (JDK removed internal API)

jdeps --jdk-internals chartfx-dataset/target/chartfx-dataset-11.1.0-SNAPSHOT.jar 
chartfx-dataset-11.1.0-SNAPSHOT.jar -> jdk.unsupported
   de.gsi.dataset.serializer.spi.FastByteBuffer       -> sun.misc.Unsafe                                    JDK internal API (jdk.unsupported)

JDK Internal API                         Suggested Replacement
----------------                         ---------------------
sun.misc.Unsafe                          See http://openjdk.java.net/jeps/260
```

Unsafe replacement links to [Variable Handles](http://openjdk.java.net/jeps/193).

 - [x] com.sun.javafx.perf.PerformanceTracker (chartfx-chart: SimplePerformanceMeter)
 - [x] sun.misc.Unsafe (chartfx-datasets: FastByteBuffer)

## unnamed packages

> Your library can't have classes in the default (unnamed) package. This is a bad idea regardless, but when your library is used as automatic module, this rule is enforced by the module system.

  - [x] no problems here

## split packages

[...]

PR #538, opened 2022-07-26
Fix to prevent illegal-access-error proposed in https://github.com/fair-acc/chart-fx/issues/537

miho:
Fix to prevent illegal-access-error proposed in https://github.com/fair-acc/chart-fx/issues/537


wirew0rm (approved):
Looks good, sorry for the long delay.

Some of the CI is also acting weird, but with these minimal changes I think we can safely ignore the missing checks.

The modularization topic in general is a bit of a sore spot, I would love to have chart-fx fully modularized #160, but for libraries which have some unmodularized dependencies there is no viable migration strategy (the closest thing is repackaging the affected dependencies in our own build, which would bring its own problems and would be quite the effort to implement correctly). And the returns for that work seem to be minimal or of a more theoretical nature at least at the moment (or badly communicated).

Issue #539, opened 2022-08-19
Desktop scaling is not respected


MarsBlessed:
**Describe the bug**
When desktop scaling is not 100% things start to behave unexpectedly. For example, `SquareButton` starts forcing `ToolBar` height to grow on each pulse of layout rendering. Another example is `DefaultNumericAxis.getDisplayPosition` which fails `de.gsi.chart.plugins.DataPointTooltipTest` due to wrong coordinates to move the desktop pointer to.

**To Reproduce**
In the wild: scale up the desktop icons to say 125% then run `de.gsi.chart.samples.DataViewerSample` and click on "the thermometer" - the toolbar will start to grow indefinitely 
In the tests: add the following lines to the end of `de.gsi.chart.plugins.DataPointTooltipTest.start` method and run the test 
``` java
        scene.getWindow().setRenderScaleY(1.25);
        scene.getWindow().setRenderScaleX(1.25);
```

**Environment:**
 - OS: Windows 7
 - Java version: temurin-11.0.16
 - JavaFx version: 13
 - ChartFx version: 11.2.7

MarsBlessed:
I have a simple solution for `SquareButton` but it seems to be just a tip of an iceberg.

wirew0rm:
That's interesting, until now I didn't have any HiDPI setup where this would cause problems, but of course this nowadays is a pretty standard thing which has to be supported. I'll have to look into how this is generally handled in javafx.

Thanks for sharing your findings and providing a way to test this. Please feel free to open a (draft - if you think it is not ready to be merged) PR for your SquareButton solution even though I fear that you are correct in your iceberg assesment.

MarsBlessed:
> Please feel free to open a (draft - if you think it is not ready to be merged) PR

here it is: [#541](https://github.com/fair-acc/chart-fx/pull/541)
>  I'll have to look into how this is generally handled in javafx

I could share some of my understanding if you'd like

MarsBlessed:
the PR seems ok to merge now but as the logs showed the merge will be prevented by CI which requires the build to run successfully using JDK 15 - this is another issue which is beyond the scope of my efforts for the bug

PR #541, opened 2022-08-22
Issue #539: SquareButton fix for scaling issue with test

MarsBlessed:


MarsBlessed:
it seems like it is not yet ready to be merged - I see that I forgot to test the code with non-zero margin and padding, I will make another commit a bit later

MarsBlessed:
the PR seems ok to merge now but as the logs showed the merge will be prevented by CI which requires the build to run successfully using JDK 15 - this is another issue which is beyond the scope of my efforts for the bug

wirew0rm:
Hey thanks for the PR and sorry for my slow response. I'm not able to reproduce the bugs on my systems using `-Dglass.gtk.uiScale=125%` (the windows eqivalent woudl be `-Dglass.win.uiScale=125%`). The `setRenderScaleX/Y()` method did not work for me, it seems to be important when in the scene's lifecycle this is called. I tried most of the samples and couldn't observe the evergrowing toolbar and for the test the only one failing is `WriteFxImageTests::testWritingImageByteBufferPalette`, while i could observe the scaled UI, so I'm confident it used the scaling. If you have some more detailed instructions how to reproduce the bugs I'll recheck, but so far this looks like a windows only bug.

This still makes this a valid problem but it means that i can only look at the code and otherwise must rely on your observations as I don't have access to a windows dev environment at the moment.

If I understand the changes correctly, the trick is to ensure that the sizes are always set to a value that will be an integer value after scaling? 

> the PR seems ok to merge now but as the logs showed the merge will be prevented by CI which requires the build to run successfully using JDK 15 - this is another issue which is beyond the scope of my efforts for the bug

The JDK15 test really is optional and won't block the merge. I also expect the initial jdk15 build error to be a flake which will go away after a rebuild. I don't see a reason why this would be jdk dependent. However there seem to be some differences in the scaling code depending on the platform. The headless (monocle) testing in the CI [fails](https://github.com/fair-acc/chart-fx/runs/8017244277?check_suite_focus=true#step:7:2794) for jdk11, too. Locally on my Linux system I get the same behavior.

It seems that GitHub actions now requires a maintainer to sign off on any CI runs, that's why the build status is not directly updated. Feel free to ping to run the CI, I'll investigate if we can restore the old behavior of just running the ci directly.

Could you take a look at the linked CI logs to see whether you have an idea why it differs from when you run the tests locally?

MarsBlessed:
>  I'm not able to reproduce the bugs on my systems using -Dglass.gtk.uiScale=125%

sorry to hear that, tried both myself and both worked exactly as I described. Thus most likely you are right and the issue it tightly coupled with the native window rendering mechanisms
> the trick is to ensure that the sizes are always set to a value that will be an integer value after scaling?

not exactly; the whole story is about `Region::isSnapToPixel` and `Region::scaleSizeY`; if you look inside of the latter, you'll find out that when snapToPixel is true (which is the default) then sizes are adjusted using `Math::ceil` function in order to find the lowest scaled sized in whole pixels which includes the current element size; the result of the adjustment is a fraction added to the original element size - in case of `ToolBar` it is the height of the its `HBox` which is the parent for a `SquareButton`. So when that happened, during the next layout pass one of the set-up listeners got called and `SquareButton::adjustDimension` is executed and during that execution `parent.getHeight` returns the adjusted value and the button adds the fraction to its preferred height. And here we go - the toolbar sees that one of its child becomes bigger, adjusts itself accordingly and force the whole cycle again - for some unclear reason scaling always results in another fraction added and thus the growing becomes self-sustained. 
My solution uses `Math::floor` to find the biggest scaled button height + paddings in whole pixels which is lower than the available size of the button parent

MarsBlessed:
> Could you take a look at the linked CI logs to see whether you have an idea why it differs from when you run the tests locally?

oh, I think I know exactly why it differs - I set the button padding in mm which could be different in different systems when expressed in pixels :) So the test works fine on my laptop but I have no idea what value would the size results in on the building machine,, Let me try to fix the test in that matter

MarsBlessed:
this also means that my tests are not perfect when it comes to different margins and paddings but making them perfect looks like overkill

MarsBlessed:
it looks like it didn't help, so the test issue could be actually the main code issue
and my main question now is whether a preferred size of a region includes its paddings or not

MarsBlessed:
unbelievable! seams like Windows allows the height of a region to be a decimal with fraction, while unbuntu only allows integers - that can explain why you couldn't reproduce the issue

MarsBlessed:
I still can't explain why the resulting height of the SquareButton in a ContextMenu is 35 on Ubuntu and 34 on Windows (for scaling of 1.0) - don't know what constitutes to that calculation apart from margin and padding. After all, I adjusted the delta for assertions according to the scaling, which should be reliable for any build environment

wirew0rm (approved):
This looks good, thanks for all your efforts in fixing the tests and sorry for not merging this sooner, it seems I missed the fixed tests due to the failing deploy step  (unrelated to the PR).

PR #557, opened 2022-11-06
resolve performance issue caused by using DoubleArrayCache.getArrayExact

protogenes:
displaying live data from data sets with fluctuating size created a lot of GC pressure because the DoubleArrayCache required a new allocation almost every call to getArrayExact

adjusting the ErrorDataSetRenderer and related classes to call DoubleArrayCache.getArray (using best fit strategy) instead removes this overhead

the time spend in the cache easily took 90% of the render time in our case, because the cache list grews quite large

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/transforms/DefaultAxisTransform.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/transforms/DefaultAxisTransform.java
@@ -25,12 +25,12 @@ public class DefaultAxisTransform extends AbstractAxisTransform {
 
     @Override
     public double getRoundedMaximumRange(final double max) {
-        return Math.floor(max);
+        return Math.ceil(max);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


ennerf:
@RalphSteinhagen I remember having a discussion about this a couple years ago, but I can't find it anymore. I think there was a reason why you wanted to pick exact sized arrays, and we ended up discussing some plugin approach where users could provide a custom allocator/object-pool.

I found an initial proof of concept ([here](https://github.com/ennerf/chart-fx/tree/ennerf/poc-array-allocation)), but I can't remember whether the idea went anywhere.

RalphSteinhagen:
Thanks @ennerf for finding and linking to the old PR and discussion thread. I guess most of what has been described there still holds true.

The core question is a 80/20-type decision: default/safe behaviour for novice/new users (the 80% use case) vs. extensibility for more custom/special cases (the 20% use case). 

We are open for PR. May I suggest: rather than modifying the existing code (which may perturb other users) it would be nice to generate an interface and allow to supply of custom array caches. 

@protogenes and/or @ennerf would you be up for it?

PR #575, opened 2023-03-09
Label refactoring

ennerf:
This PR removes the 20MB icu4j dependency and implements a custom number formatter based on Schubfach. It also removes the temporary string conversions from `DigitNumberArithmetic`. The format changed a bit to make labels easier to read. Here is a video of the currently implemented behavior: https://youtu.be/Od5KhBZrWiU.

**Open questions**
- [x] The `DefaultFormatter` now handles log cases as well. For now I added a deprecated annotation to `DefaultLogFormatter`, but maybe it should be deleted?
- [x] The `DefaultFormatter::fromString` used the format for parsing. Did that do anything special that does not work with `Double::parse`?
- [x] What is the `rangeIndex` variable in `DefaultFormatter`? I didn't see any uses of it.
- [x] The default formatter interprets `NumberFormatter::setPrecision` as the number of after comma digits plus one. Is this reasonable or should we change the name to `setAfterCommaDigits` and potentially break backwards compatibility?
- [x] Should the exponential form use a lower case `e` to make it easier to distinguish between significand and exponent? e.g.

image
image
image image

**TODOs**
- [ ] Remove commented out debug prints

**Behavior differences**

| Before | After | Comment     |
|--------|-------|-------------|
| image | image | unnecessary after comma zeros are removed |
| image  | image | exponential axes render as plain numbers if all ticks are within the plain format range |
| image | image |  labels are rounded to the first significant digit  |
| image |image| plain labels keep their form and just add extra digits |

**Known issues**
| Before | After | Comment     |
|--------|-------|-------------|
| image | image | The zero values in the exponential format can be tiny and stick out (same behavior, so not a regression). Sometimes zero gets rendered as negative zero. The plain format has a workaround, but the easiest place to fix this may be in the tick renderer. |


chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/format/DefaultFormatter.java
```diff
@@ -77,84 +66,142 @@ public DefaultFormatter(final Axis axis, final String prefix, final String suffi
...
+            configureFormatter(getRange(), majorTickMarksCopy);
+
+            // Clear the cache if the formatting changed
+            if (formatter.isExponentialForm() != prevExponentialForm
+                    || formatter.getPrecision() != prevPrecision
+                    || oldRangeIndex != rangeIndex) {
+                labelCache.clear();
+
+                // TODO: what is the rangeIndex? it does not get set anywhere. Maybe that's a subclass thing?
+                oldRangeIndex = rangeIndex;
```
wirew0rm, line 93: Instead of saving the old state of all properties, we could just increment range index in the setters of the properties. But then again we could also call clear there directly (clearing an already cleared cache shouldn't be expensive).

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/format/DefaultLogFormatter.java
```diff
@@ -12,6 +12,7 @@
 /**
  * @author rstein
  */
+@Deprecated // supported by DefaultFormatter
```
wirew0rm, line 15: IMO its safe to delete it if its functionality is subsumed in the DefaultFormatter. The only usecase would be as an example to implement custom Formatters. If there is some feature of the old implementation that users rely on it's still possible to incorporate the old version and reintroduce it if needed or add the functionality to your new Formatter.

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/format/DefaultFormatter.java
```diff
@@ -77,84 +66,142 @@ public DefaultFormatter(final Axis axis, final String prefix, final String suffi
...
+        int minExp = decompositions.get(0).getExponent();
+        int maxExp = minExp;
+        for (int i = 1; i < n; i++) {
+            final int exp = decompositions.get(i).getExponent();
+            minExp = Math.min(minExp, exp);
+            maxExp = Math.max(maxExp, exp);
+        }
+
+        // Use the exponential form for very large or very small numbers
+        boolean useExponentialForm = minExp < -3 || maxExp > 4;
```
wirew0rm, line 138: Would it make sense to have these configurable? This would also allow to force exponential form (min/maxExp=0) or decimal form (min/maxExp = MIN/MAX_INT).
Also a third condition could be the difference between min and max exponent: If you have labels with 1.5e-2 and 1.5e+3 its easier to read than 0.15 and 1500.00? I think the way the tickmarks are generated this should only be triggered by log axes and if we make it configurable users can select if they want the before or after behavior from your screenshots.

chartfx-chart/src/test/java/io/fair_acc/chartfx/utils/NumberFormatterImplTest.java
```diff
@@ -0,0 +1,120 @@
...
+    void plainFormatEnglish() {
+        Locale.setDefault(Locale.US);
+        var formatter = createFormatter(false, ALL_DIGITS);
+        assertEquals("0", formatter.apply(0));
+        assertEquals("2", formatter.apply(2.0));
+        assertEquals("2.1", formatter.apply(2.1));
+        assertEquals("10", formatter.apply(10));
+        assertEquals("123.456789", formatter.apply(123.456789));
+        assertEquals("1.23456789E14", formatter.apply(123.456789E12));
+        assertEquals("0.000000000123456789", formatter.apply(123.456789E-12));
```
wirew0rm, line 43: It bugs me a bit that the behavior is so different for positive and negative exponents, intuitively I would expect them to behave in a symmetric way. Is there an easy rationale to why it has to be that way?
I realize that these are pathological cases that will never be triggered in the label code due to the way the labels are generated, but some plugins or users might want to do number formatting consistent with the labels and feed the formatter arbitrary data (e.g. the value indicators).

wirew0rm (commented):
> This PR removes the 20MB icu4j dependency and implements a custom number formatter based on Schubfach. It also removes the temporary string conversions from `DigitNumberArithmetic`. The format changed a bit to make labels easier to read. Here is a video of the currently implemented behavior: https://youtu.be/Od5KhBZrWiU.

Very nice work, thanks for tackling this, looks very good. And special thanks for documenting all the changes and open questions so nicely!

I tried to answer the questions directly in the code when applicable to the best of my understanding, but I 

> Should the exponential form use a lower case `e` to make it easier to distinguish between significand and exponent? e.g.

I agree that it's more readable, but outside of programming I think it is avoided to prevent mistaking it with the fundamental constant. This is not a problem for axes, but it is not uncommon to have a UI which displays a chart and formula side by side and one would want to have consistent number representation between them.  `1,23456789ᴇ14` is another variant i've seen in the wild  and there even is a dedicated unicode symbol `6.022⏨23` but I think it is not that common.
It looks like this is configurable via the locale dependent `java.text.DecimalFormatSymbols`.
I'm also looking into making more things Stylable properties, so we could have something like `.chart {scientific-notation-separator: "e";}` in the user css. Out of scope for this PR.

> **Behavior differences**

I agree with all of them being clear improvements except for the logarithmic axis, where people might actually prefer the default of the exponential notation.

> **Known issues**
> Before 	After 	Comment
> image 	image 	The zero values in the exponential format can be tiny and stick out (same behavior, so not a regression). Sometimes zero gets rendered as negative zero. The plain format has a workaround, but the easiest place to fix this may be in the tick renderer.

This would mean a special case in the renderer to render a zero label if the value is very small compared to the range?

chartfx-chart/src/test/java/io/fair_acc/chartfx/utils/NumberFormatterImplTest.java
ennerf: The byte buffer has a limited size that would not be able to fit very large numbers (e.g. Double.MAX_VALUE would need 300 zero digits). It took me too long to figure out the special casing, and I didn't want to waste buffer space for something that'll never happen, so I added an arbitrary threshold at which it just uses the exponential form.

I originally had a similar threshold for negative exponents, but I added the required special casing to support zooming around zero without changing formats.

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/format/DefaultFormatter.java
ennerf: My personal preference is actually for `0.15` and `1500`. I added a `DefaultFormatter::setDisplayFormat(enum)` with `Auto`, `Scientific`, and `Plain` options to force a specific format. I also moved the range check into a protected method for the rare use cases where users may want to override the auto behavior.

ennerf:
> This would mean a special case in the renderer to render a zero label if the value is very small compared to the range?

I think so. It's difficult to back out the zero at a lower point and forward the information to the formatter.

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/format/DefaultFormatter.java
ennerf: for now I left those lines as they were

ennerf:
I didn't like the rounding of `0.25` steps to `[0.0, 0.3, 0.5, 0.8, 1.0]`, so I added an extra digit for cases where the next digit is not zero.

image   image  image

wirew0rm (approved):
Thanks for all the work you put into this, looks and feels great, lets get this merged.

PR #576, opened 2023-03-26
remove fastutil dependency

ennerf:
`dataset` currently depends on `it.unimi.dsi:fastutil`, which adds a ~19MB dependency for 3 classes that are actually being used. After icu4j the total size with dependencies (other than JavaFX) is about ~29MB, so removing it would reduce the total size by 2/3 to below 10MB.

> mvn dependency:copy-dependencies -DincludeScope=runtime
> ls -lhS chartfx-chart/target/dependency

```
 19M fastutil-8.3.1.jar
2.2M commons-math3-3.6.1.jar
1.4M controlsfx-11.1.1.jar
1.2M JTransforms-3.1.jar
574K commons-lang3-3.12.0.jar
433K math-master-SNAPSHOT.jar
387K dataset-master-SNAPSHOT.jar
348K ikonli-fontawesome5-pack-11.5.0.jar
228K JLargeArrays-1.5.jar
164K pngj-2.1.0.jar
156K ikonli-fontawesome-pack-11.5.0.jar
 51K slf4j-api-2.0.0-alpha0.jar
 37K ikonli-javafx-11.5.0.jar
 27K annotations-21.0.1.jar
 11K ikonli-core-11.5.0.jar
```

In this PR I shaded the fastutil classes and removed some parts that aren't being used (e.g. sorting). Fastutil uses the Apache v2 license, so it could be used as is.

Do you think it'd be worth using the generator to create more custom-fit implementations?

### DIFF
--- a/chartfx-dataset/pom.xml
+++ b/chartfx-dataset/pom.xml
@@ -20,11 +20,6 @@
 	</description>
 
     <dependencies>
-        <dependency>
-            <groupId>it.unimi.dsi</groupId>
-            <artifactId>fastutil</artifactId>
-            <version>8.3.1</version>
-        </dependency>
         <!-- micro-benchmarking framework -->
         <dependency>
             <groupId>org.openjdk.jmh</groupId>
--- a/chartfx-dataset/src/main/java/io/fair_acc/dataset/spi/DoubleDataSet.java
+++ b/chartfx-dataset/src/main/java/io/fair_acc/dataset/spi/DoubleDataSet.java
@@ -9,7 +9,7 @@ import io.fair_acc.dataset.DataSet;
 import io.fair_acc.dataset.DataSet2D;
 import io.fair_acc.dataset.EditableDataSet;
 
-import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
+import io.fair_acc.dataset.spi.fastutil.DoubleArrayList;
 
 /**
  * Implementation of the {@code DataSet} interface which stores x,y values in two separate arrays. It provides methods
--- a/chartfx-dataset/src/main/java/io/fair_acc/dataset/spi/DoubleErrorDataSet.java
+++ b/chartfx-dataset/src/main/java/io/fair_acc/dataset/spi/DoubleErrorDataSet.java
@@ -10,7 +10,7 @@ import io.fair_acc.dataset.DataSet2D;
 import io.fair_acc.dataset.DataSetError;
 import io.fair_acc.dataset.EditableDataSet;
 
-import it.unimi.dsi.fastutil.doubles.DoubleArrayList;
+import io.fair_acc.dataset.spi.fastutil.DoubleArrayList;
 
 /**
  * Implementation of the {@code DataSetError} interface which stores x,y, +eyn, and -eyn values in separate double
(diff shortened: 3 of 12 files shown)


wirew0rm:
Thanks for giving this another go.

> `dataset` currently depends on `it.unimi.dsi:fastutil`, which adds a ~19MB dependency for 3 classes that are actually being used. After icu4j the total size with dependencies (other than JavaFX) is about ~29MB, so removing it would reduce the total size by 2/3 to below 10MB.

I can see that this would be a huge improvement for applications where the size of the resulting jar matters :+1: .

> In this PR I shaded the fastutil classes and removed some parts that aren't being used (e.g. sorting). Fastutil uses the Apache v2 license, so it could be used as is.

Any reason to use manual shading and not rely on the [shading plugin](https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html)? Benefits would be the ability to follow updates (fastutil seems reasonably stable, but it's still actively maintained), document/reduce code ownership, making it easy to use other classes from fastutil in the future in a consistent way. I haven't used that functionality myself before, so I might overlook some complications.

Another solution that gets mentioned in the fastutils issue is using proguard to produce a reduced jar, but I'd rather not make all users depend on a non-standard solution.

> Do you think it'd be worth using the generator to create more custom-fit implementations?

That would essentially mean re-implementing fastutil (or the small subset of it that we are using) with our own templating (fastutil is generated using a C-preprocessor based code-generator), right? Do you mean custom-fit in terms of datatypes or in terms of features/performance characteristics? I suspect you have specific improvements in mind?
I think it's probably feasible with the infrastructure we have in place, but since this type of "custom templating" code is harder to read/modify/maintain I think it should only be done if it there are substantial benefits over vendoring fastutil (independent of if it is done via having an explicit copy in our tree or programatically from maven).

In general this looks good, and I wouldn't be opposed to merging this, either because there are aguments against sharding in the maven step that i'm not aware of or simply because noone has time to implement it and we have a working solution to a real problem here.

ennerf:
> Any reason to use manual shading and not rely on the [shading plugin](https://maven.apache.org/plugins/maven-shade-plugin/examples/class-relocation.html)? Benefits would be the ability to follow updates (fastutil seems reasonably stable, but it's still actively maintained), document/reduce code ownership, making it easy to use other classes from fastutil in the future in a consistent way. I haven't used that functionality myself before, so I might overlook some complications.

Using the shading plugin would pull in a lot of other classes like `DoubleCollection`, `DoubleList`, and various utility classes for unused features like iterators and sorting. I think it'd be a decent amount of effort to find an appropriate subset of classes that still works.

It was fairly straight forward to get the lists in, so I don't think adding more classes (e.g. a primitive map) would be that difficult to add later if needed. The basic features of primitive lists are incredibly stable, so I don't expect any significant updates that would need to be backported.

> that would essentially mean re-implementing fastutil (or the small subset of it that we are using) with our own templating (fastutil is generated using a C-preprocessor based code-generator), right?

Yes, I was mostly thinking of naming conventions for consistency. There is an argument for keeping it as is though to simplify pulling in more features if needed.

wirew0rm (approved):
Looks good. I added a small notice to the top of the copied files linking to the original project and stating that the files where modified/simplified.

PR #580, opened 2023-04-06
added conveyor packaging for fx-sampler

ennerf:
I added a basic [Conveyor ](https://conveyor.hydraulic.dev/) configuration for the sampler app and uploaded a signed test version to [chartfx-sampler](https://files.hebi.us/internal/demo/chartfx-sampler/download.html). I also added a preliminary square icon to match the macOS guidelines, and fixed a few issues that broke the sampler app when running from jar files (e.g. resource paths with `.` and `..` don't work)

It probably doesn't make sense to get signing certificates for a sampler project. Maybe Hydraulic (ping @mikehearn) can handle the signing/hosting like they do for some other projects?

I'm leaving it as a draft PR for now because there is some issue with the installed versions. See [conveyor#56](https://github.com/hydraulic-software/conveyor/issues/56)

**Update**
The issue occurs when executing with `bin` as the working directory. Executing it manually from somewhere else works. Absolute vs relative paths do not make a difference. My best guess is that it's a bug in fx-sampler.


ennerf:
The Conveyor packaging format isn't compatible with the FXSampler lookup, but I managed to work around it. Everything should be working now:

Test release: https://files.hebi.us/internal/demo/chartfx-sampler/download.html

ennerf:
I built a new version with the new logos and 11.3 changes: [download link](https://hebifiles.z13.web.core.windows.net/internal/demo/chartfx-sampler/download.html)

wirew0rm (approved):
Looks good.

wirew0rm (approved):
:+1:

PR #581, opened 2023-04-10
Fix sample data paths

uosis:
Broken by https://github.com/fair-acc/chart-fx/pull/578 it seems.


wirew0rm (approved):
Thanks for catching this and contributing back :+1:

Issue #579, opened 2023-04-06
[10pt, 10pt]Restructure layout computations to make rendered output more consistent


wirew0rm:
Goals:
- ensure chart is always self-consistent after a rendering pass
  - having a single frame with e.g. some clipping is acceptable for a frame, if otherwise we would have to make very pessimistic assumptions which would be bad for performance, but all chart elements (labels, grid, plots) should be at consistent positions in every frame.
- reduce the levels of scene graph hierarchy of the `Chart` component
- ideally this will solve or make the solution a lot simpler for these issues: #474, #410, #204
- improve integration with JavaFx properties and CSS styling facilities

Items to be investigated implemented:
- Separate range computations, layout calculations and actual graphics context rendering in the layout phase by
  - either calling different functions in the layoutChildren implementation of Chart
  - or using the pre-/post-render hooks of javafx
- Reimplement Chart as a layout container which explicitly implements a custom layout Children
  - preiminary investigation/implementation in #529 and the corresponding newLayout branch
- implement global locking of datasets per render pass
  - currently range computation and rendering separately lock the dataset which is not safe and requires multiple checks
- have a dataset node type to hold things like dirty bit, bindable properties, etc
  - datasets can update stylable properties of the wrapper via their style (or add style classes)    
  - investigate relationship of this new wrapper to DataSetManagement (base class of renderer)
  - chart (and renderers?) store List<DataSetNode>, helper functions to get or remove raw DataSet
  - DataSetNode can be in scenegraph once, but Multiple DataSetNodes can refer to the same DataSet
  - DataSetNode is only handled in FxThread, DataSet is thread safe by default
- have a simple way to disable multithreading in the DataSet
  - should default to safe/locked implementation

ennerf:
As per discussion on June 21 with Alex. Initial steps:

- [ ] Split layout and rendering
- [ ] Perform layout based on properties in the children nodes

- [ ] Remove XYChart class and use Chart directly
- [ ] Remove dataset list from main chart and keep datasets within the renderer. The global color index should be maintained.
- [ ] Deprecate delegating methods, e.g., chart delegating to the grid renderer
- [ ] Change renderer to always have one X and one Y axis. If one is not defined it defaults to the first one.

ennerf:
I think the goals have all been achieved and are ready for review. The style nodes ended up not being necessary.

ennerf:
Below are some before and after pictures

| Before | After | Comment     |
|--------|-------|-------------|
[...]

Issue #530, opened 2022-06-29
new LMAX Disruptor-based event notification system


wirew0rm:

ennerf:
Matching the ideas I've mentioned before, I created an event framework based on bit masks where listeners can subscribe to dirty bits that they are interested in. Similar to JavaFX properties, each element can subscribe to others via change-listeners (called if specified bits change from 0 to 1) or invalidation-listeners (called on every event). 

Example for an axis that needs to trigger a layout on a property change

```Java
// create an axis state object that knows only about axis events
var state = BitState.initDirty(this, ChartBits.AxisMask);

// changes to the axis padding need to recompute the layout and redraw the canvas
axisPadding.addListener(state .onPropChange(ChartBits.AxisLayout, ChartBits.AxisCanvas)::set);

// trigger JavaFX layouts
state.addChangeListener(ChartBits.AxisLayout, (src, bits) -> requestLayout());
```

Example for a chart that subscribes and unsubscribes to an axis:

```Java
// create a chart state object that knows about all events
var state = BitState.initDirty(this, ChartBits.AxisMask);

// merge changes coming from a relevant axis
axis.getBitState().addChangeListener(this.state);

// trigger a redraw if any axis needs to be drawn
state.addChangeListener(ChartBits.AxisCanvas, (src, bits) -> drawAxesInNextCycle());

// remove an axis that is no longer part of this chart
axis.getBitState().removeChangeListener(this.state);
```

Example for skipping the draw step if none of the axes has changed
```Java
redrawAxes() {
  if (state.isClean(ChartBits.AxisCanvas) {
    return; // all content is still good
  }
  for (var axis : getAxes) {
    axis.redraw();
  }
}

```

The state bits get cleared after all elements are drawn. Updates use bitwise operations and batch updates automatically, so the process is very efficient. For example 10 data sets changing 100 values will get merged into 10 dataset events that are then merged into a single chart event.

The merging/batching also simplifies scenarios where datasets are updated from non-JavaFX threads, e.g., with a simple `runLater()` handoff we get a maximum of one runLater call per non-JavaFX thread per dataset, e.g.,

```Java
dataSet.setDirty(ChartBits.Data, CharBits.DataRange);
dataSet.getBitState().addChangeListener(this.state, (src, bits) -> runOnFx(() -> this.state.setDirty(bits));
```

Depending on the needs we could also have variants with thread-local state and/or smarter locking.

[...]

PR #592, opened 2023-07-21
Deterministic layout

ennerf:
PR for #579. The layout and rendering was significantly simplified. There are still some feedback loops and iterative renderings, but that should be fixed later with #530.


wirew0rm (approved):
Really great work @ennerf, this greatly reduces the rendering inconsistencies between charts and axes and also reduces the complexity which made it increasingly harder to follow up on issues concerning layout and axes.
The most important improvements are demonstrated in: https://github.com/fair-acc/chart-fx/issues/579#issuecomment-1661238946

There are still some issues after this PR where the problems overlap with the follow-up notification system which are fixed there.
Formatting will be done as a tree wide change including upgrading the clang-format version before the release.

Overall the changes greatly improve reliability and maintainability of the library enabling continued maintenance to follow up on smaller issues or introducing new features. The changes are also nicely structured and documented.

Thanks a lot for all the hard work that was put into this!

PR #594, opened 2023-08-02
Layout and notification system

ennerf:
Extension on PR #592 that fixes issues #579 and #530. See issues for more information.


wirew0rm (dismissed):
Very nicely designed event system overhaul. The architecture (as described in [Architecture.md](https://github.com/fair-acc/chart-fx/blob/7faf6e5a5d9a8e2ba6b54a185c7c3417016d5134/docs/Architecture.md)) looks to be very efficient at ensuring that everything is exactly coputed/rendered one time, as the [results](https://github.com/fair-acc/chart-fx/issues/530#issuecomment-1661303122) prove. This improves performance, snappiness and also reliablity of things like textfx unit-tests or use-cases that generate screenshots from charts where the chart is not constantly updated but only rendered on demand.

There remains some work on (multithreaded) processing of dataset events, as outlined in https://github.com/fair-acc/chart-fx/issues/530#issuecomment-1660041168 which has to be tackled before the release.

Thanks for the all the work in bringing this into chartfx, I believe the library will benefit greatly from the new system in the future.

wirew0rm (approved):
Reapproving because resolving the merge conflicts removed the approval, sorry for the noise.

PR #598, opened 2023-08-10
Axis and renderer CSS Styling

### LINKED ISSUE #560: Fix CSS styling for GridRenderer
**Describe the bug**

chart.css lists CSS selectors for styling the grid lines, but overwriting them currently does not work.

```
.chart-major-vertical-lines .chart-major-grid-lines {
  /* use this to override v-specific settings */
}

.chart-major-horizontal-grid-lines .chart-major-grid-lines {
  /* use this to override h-specific settings */
}
```

This is because `GridRenderer` lives in a separate Scene with manual updates to `applyCss`, so it does not see any styles applied to the main chart. Applying styles directly to the renderer should work, but somehow that seems to have issues too.

**Workaround**

For now I made it work by adding the GridRenderer to an invisible/unmanaged group to an unused panel, e.g.,

```Java
public static <T extends Chart> T fixChartCss(T chart) {
    if (chart instanceof XYChart xy) { 
        var hiddenGroup = new Group(xy.getGridRenderer());
        hiddenGroup.setVisible(false);
        hiddenGroup.setManaged(false);
        xy.getToolBar().getChildren().add(hiddenGroup);
    }
    return chart;
}
```

**Discussion**

The approach with invisible lines seems to work well, but I think they should be part of the main scenegraph to avoid special casing and odd behavior. Do you think it would make sense to add a dedicated non-visible/unmanaged portion in `Chart` that exists only for CSS updates?

If so, how should the GridRenderer register to the chart? The quick hack would be to add it on `render(chart)`, but it may make sense to register it in a similar manner as plugins.

ennerf:
This PR includes some styling changes for the components that are comparatively easy to deal with. Datasets will require more work and will be in a separate PR.

* added SCSS integration
* made axes fully styleable (including the side)
* fixed CSS styling for the grid renderer
* added CSS styling for the error renderer
* removed duplicate properties
* removed many convenience methods for styling, e.g., setLabelFill(color)` -> `getLabel.setFill(color)`. This may result in a bit more effort to port, but it cleans up the API and should help encourage the use of CSS.
* added extra pseudo classes for sides (e.g. `vertical`, `horizontal`, `center`)

**Bugfixes**
* Fixes #560 grid styling
* fixed some tick-label artifacts caused by rendering labels outside the parent pane (e.g. the bottom-left x-axis tick if the y axis is on the right)

**Demos with Sass and live CSS reloading**

* [ChartFX#598: Renderer styling demo](https://youtu.be/ExJ0gswwjd0)
* [ChartFX#598: Axis styling demo](https://youtu.be/9LsBp9vCVKU)

**Notes**

- [ ] The CssPropertyFactory in some Axes (e.g. `NumericAxis`) gets initialized with the static properties from the parent class before they are valid. This is a bug in current main.
- [ ] The legend initializes and renders the symbols before the CSS info is available. The symbols should be updated in the draw phase.

### DIFF
--- a/chartfx-chart/pom.xml
+++ b/chartfx-chart/pom.xml
@@ -13,6 +13,9 @@
     <name>chartfx-chart</name>
     <properties>
         <project.moduleName>io.fair-acc.chartfx</project.moduleName>
+        <sass.version>1.64.2</sass.version>
+        <scss.inputDir>${project.basedir}/src/main/resources/io/fair_acc/chartfx/</scss.inputDir>
+        <css.outputDir>${scss.inputDir}</css.outputDir>
     </properties>
 
     <description>This charting library ${project.artifactId}- is an extension
@@ -93,5 +96,32 @@
             <version>2.1.0</version>
         </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+            <plugin> <!-- Converts Sass files to CSS -->
+                <groupId>us.hebi.sass</groupId>
+                <artifactId>sass-cli-maven-plugin</artifactId>
+                <version>1.0.3</version>
+                <configuration>
+                    <sassVersion>${sass.version}</sassVersion>
+                    <args>
+                        <arg>${scss.inputDir}:${css.outputDir}</arg>
+                        <arg>--no-source-map</arg>
+                    </args>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>sass-exec</id>
+                        <phase>generate-resources</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
 </project>
 
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxis.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxis.java
@@ -40,6 +40,7 @@ public abstract class AbstractAxis extends AbstractAxisParameter implements Axis
     protected static final double MAX_NARROW_FONT_SCALE = 1.0;
     protected static final double MIN_TICK_GAP = 1.0;
     private final transient Canvas canvas = new ResizableCanvas();
+    private boolean drawAxisLabel;
     private boolean shiftLabels;
     protected boolean labelOverlap;
     protected double scaleFont = 1.0;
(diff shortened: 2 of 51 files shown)


chartfx-chart/pom.xml
```diff
@@ -13,6 +13,9 @@
     <name>chartfx-chart</name>
     <properties>
         <project.moduleName>io.fair-acc.chartfx</project.moduleName>
+        <sass.version>1.64.2</sass.version>
+        <scss.inputDir>${project.basedir}/src/main/resources/io/fair_acc/chartfx/</scss.inputDir>
+        <css.outputDir>${scss.inputDir}</css.outputDir>
```
wirew0rm, line 18:
``` diff
-  <scss.inputDir>${project.basedir}/src/main/resources/io/fair_acc/chartfx/</scss.inputDir>
- <css.outputDir>${scss.inputDir}</css.outputDir>
+ <scss.inputDir>${project.basedir}/src/main/scss/</scss.inputDir>
+ <scss.outputDir>${project.basedir}/src/main/resources/io/fair_acc/chartfx/</scss.outputDir>
```
If` I'm not misstaken this leads to both the scss and css files being copied into the resulting jar, while only the generated css files are required. Not a big deal, but it might be cleaner to have them in a separate directory. As I don't have a lot of experience with scss I'm not sure how this all plays together.

Also the idea is to have scss and the generated css files in the git repo?

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxis.java
```diff
@@ -598,6 +598,43 @@ private void applyOverlapPolicy(List<TickMark> tickMarks) {
...
     }
 
+    /**
+     * Hides labels that can't be fully displayed due to the
+     * bounds of the parent container. This often manifests
+     * on the top of y axes when there is no title padding.
+     * <p>
+     * Note that the layout bounds may not be available
+     * until the bounds phase, so we use layout x/y directly.
+     */
```
wirew0rm, line 608: If people want to show e.g. the rightmost tick on their x axis without having something to show on the right side of the chart, I assume they could just add some padding to the right side of the chart?

wirew0rm (dismissed):
Very nice work, looks good and can be merged once the tests are fixed.

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxis.java
ennerf: yes, they could add padding via CSS

```
.chart-content {
  -fx-padding: 5px;
}
```

Issue #571, opened 2023-02-28
Application does not exit with chartfx


PavelSpiridonov:
### Discussed in https://github.com/fair-acc/chart-fx/discussions/467

<div type='discussions-op-text'>

<sup>Originally posted by **DennisOlbrich** September 28, 2021</sup>
If any gsi XYchart is used in the FX application, the application will not close, when the last window is closed. Do you have any idea what is wrong?

`package application;

import de.gsi.chart.XYChart;
import javafx.application.Application;
import javafx.scene.Scene;
import javafx.scene.layout.AnchorPane;
import javafx.stage.Stage;

public class XYChartMain extends Application {

	@Override
	public void start(Stage primaryStage) throws Exception {
		XYChart chart = new XYChart();

		AnchorPane ap = new AnchorPane();
		ap.getChildren().add(chart);

		Scene scene = new Scene(ap, 640, 480);
		primaryStage.setScene(scene);

		primaryStage.show();
	}

	public static void main(String[] args) {
		// TODO Auto-generated method stub
		launch(args);
	}
}`</div>

milo-gsi:
Try adding 

```
primaryStage.setOnCloseRequest((e) -> {
	System.exit(0);
});
```

ennerf:
@wirew0rm It looks like this issue was fixed somewhere with the layout changes. The provided sample code now closes properly on `main`.

wirew0rm:
Going though the samples to narrow the problem down, these are still affected:
- ChartIndicatorSample
- DataViewerSample
- ErrorDataSetRendererSample
- Histogram2DimSample
- OscilloscopeAxisSample
- PaddedGrowAxis
- RollingBuffer[SortedTree]Sample
- WaterfallPerformanceSample

So that narrows it down

wirew0rm:
closed as the original Problem seems to be solved. The samples will be fixed by a followup PR

PR #600, opened 2023-08-11
Samples: fix non-terminating samples

wirew0rm:
There are mainly 3 reasons for the samples not to terminate:
- SimplePerformanceMeter (and the ProfilerInfoBox based on it) This class registers some javafx hooks which will cause the application to fail to terminate. For now it is removed from the samples, but should be repaired before merging this
- PaddedGrowAxis used a non deamonised Updateer thread
- The Midi based dataset for the Waterfall plot causes problems commented out for now but should be repaired before merge
- Histogram2DimSample called runFx from a javafx timer, after removing that, it terminates correctly

followup of #571, see there for further context

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/utils/SimplePerformanceMeter.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/utils/SimplePerformanceMeter.java
@@ -76,6 +76,7 @@ public class SimplePerformanceMeter extends Region {
             @Override
             public void handle(long now) {
                 pulseCounter.getAndIncrement();
+                updateProperties();
             }
         };
 
@@ -212,8 +213,6 @@ public class SimplePerformanceMeter extends Region {
                 frameRateAvgInternal = computeAverage(frameRateInternal, frameRateAvgInternal, alpha);
                 cpuLoadProcessAvgInternal = computeAverage(cpuLoadProcessInternal, cpuLoadProcessAvgInternal, alpha);
                 cpuLoadSystemAvgInternal = computeAverage(cpuLoadSystemInternal, cpuLoadSystemAvgInternal, alpha);
-
-                FXUtils.runFX(SimplePerformanceMeter.this::updateProperties);
             }
         }, 0, updateDuration);
         scene.addPostLayoutPulseListener(pulseListener);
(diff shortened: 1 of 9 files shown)


chartfx-chart/src/main/java/io/fair_acc/chartfx/utils/SimplePerformanceMeter.java
```diff
@@ -213,7 +214,7 @@ public void run() {
                 cpuLoadProcessAvgInternal = computeAverage(cpuLoadProcessInternal, cpuLoadProcessAvgInternal, alpha);
                 cpuLoadSystemAvgInternal = computeAverage(cpuLoadSystemInternal, cpuLoadSystemAvgInternal, alpha);
 
-                FXUtils.runFX(SimplePerformanceMeter.this::updateProperties);
+                Platform.runLater(SimplePerformanceMeter.this::updateProperties);
```
ennerf, line 217: Why not use an AnimationTimer?

chartfx-samples/src/main/java/io/fair_acc/sample/chart/PaddedAutoGrowAxisSample.java
```diff
@@ -37,7 +36,7 @@ public Node getChartPanel(Stage primaryStage) {
 
         var ds = new CircularDoubleErrorDataSet("", 150);
         chart.getDatasets().addAll(ds);
-        new Thread(() -> {
+        var thread = new Thread(() -> {
```
ennerf, line 39: `CachedDaemonThreadFactory.newThread(runnable).start()` ?

chartfx-samples/src/main/java/io/fair_acc/sample/financial/service/SimpleOhlcvReplayDataSet.java
```diff
@@ -172,7 +172,9 @@ public void setUpdatePeriod(final double updatePeriod) {
     public void start() {
         paused.set(false);
         running.set(true);
-        new Thread(getDataUpdateTask()).start();
+        var thread = new Thread(getDataUpdateTask());
```
ennerf, line 175: `CachedDaemonThreadFactory.newThread(runnable).start()` ?

ennerf (dismissed):

chartfx-chart/src/main/java/io/fair_acc/chartfx/utils/SimplePerformanceMeter.java
wirew0rm: good point, also since there already is an animation timer, that can just also do the update.

PR #601, opened 2023-08-11
Fix pulse triggers

ennerf:
Contrary to what my previous testing showed, the pre/post layout hooks do not trigger a pulse. Those must have been caused by animations or other timers inside the apps I was testing with (sampler and some individual samples).

This caused some issues where concurrent updates would correctly register the hooks, but the chart would never get updated without something else triggering a pulse (e.g. mouse click, resize, etc).

PR changes:

* The hooks are always registered, but they are skipped if all states are clean
* Updates now always trigger a layout of a style group that is unmanaged and has an empty `layoutChildren` implementation. This triggers a pulse without forcing unnecessary work, i.e., it does not trigger a layout of the parent chart.
* Fixed a bug that caused auto ranges to always be invalidated even if they were not changed (this didn't show up the way we were doing it before)
* Cosmetic cleanup of the chart constructor


wirew0rm (approved):
Looks good, thanks for following this up!

PR #602, opened 2023-08-16
Fix zoom transformations to be based on display coordinates instead of data coordinates

wirew0rm:
Previously the zoomer would perform the scroll-zoom and panning calculations in data coordinates. This is identical to pixel coordinates for linear axes, but for e.g. log axes this leads to strange behaviour since it can result in negative data values.

This changes the Pan and pan and scroll behaviour, the rectangle zoom was naturally always using screen coordinates anyway, so this should also be more consistent.

To archive this a change to AbstractAxis was necessary to update cached variables on range change.

If the axis gets changed multiple times during event handling, but the cache is only updated during layout, axes transforms from the event handlers use outdated transformations. This has the effect that e.g. panning only applies the first mouse event in each pulse and the panning is a lot slower than the actual mouse movement.

Also incorporates a small fix in the pom to use valid automatic module names, converts the ErrorDataSetRendererStylingSample to a correct FxSampler sample and fixes 2 small bugs:
- the cached value of isLogAxis in NumericAxis was never updated so it could not be switched by its property
- the Plugins have been changed to use Panes instead of Groups for Positioning, resolving a bug where relayouting during an ongoing zoom action would shift the zoom rect to the upper left corner.

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/Chart.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/Chart.java
@@ -27,7 +27,6 @@ import javafx.collections.ObservableList;
 import javafx.css.*;
 import javafx.geometry.*;
 import javafx.scene.CacheHint;
-import javafx.scene.Group;
 import javafx.scene.Node;
 import javafx.scene.canvas.Canvas;
 import javafx.scene.control.Control;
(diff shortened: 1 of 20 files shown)


chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxisParameter.java
```diff
@@ -816,7 +813,9 @@ public ObjectProperty<AxisLabelOverlapPolicy> overlapPolicyProperty() {
 
     @Override
     public boolean set(final double min, final double max) {
-        return PropUtil.set(minProp, min) | PropUtil.set(maxProp, max);
+        final boolean result = PropUtil.set(minProp, min) | PropUtil.set(maxProp, max);
+        updateCachedTransforms();
```
ennerf, line 817: This should be done by the called if needed. Please remove

chartfx-chart/src/main/java/io/fair_acc/chartfx/Chart.java
```diff
@@ -928,10 +927,10 @@ private void ensureJavaFxPulse() {
         return CSS.getCssMetaData();
     }
 
-    protected static Group createChildGroup() {
-        final Group group = new Group();
+    protected static Pane createChildGroup() {
```
ennerf, line 930: maybe call it `createUnmanagedPane()` or move it to some util class

ennerf (approved):

chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/AbstractAxisParameter.java
wirew0rm: thanks for catching, this was leftover from the other attempt of tackling this. Will change.

chartfx-chart/src/main/java/io/fair_acc/chartfx/Chart.java
wirew0rm: Will do.

PR #604, opened 2023-08-20
Better profiling

ennerf:
This PR builds on top of #603 (dataset styling)

It looks like there were some profiling capabilities in the `ProcessingProfiler`, but it seems limited to logs and I don't know whether there is any tool to visualize it.

## Interfaces

I've been thinking a bit about this, and I think I came up with a nice interface that is extremely flexible while also providing very fine grained access to items of interest. It currently lives in the dataset module, but once it's stable I'd like to turn the base into a separate project so it's usable in non-chart projects as well. It consists of 3 interfaces:

* a `Profiler` is something that creates named measurements

```Java
interface Profiler {
    DurationMeasure newDuration(String tag);
}
```

* a `DurationMeasure` can measure time by using start and stop. An appropriate clock is chosen internally, but there is also a fallback to record a raw value for cases where timestamps are passed around (e.g. in event sourcing timestamps are often kept with the events)

```Java
interface DurationMeasure {
    void start();
    void stop();

   TimeUnit getClockUnit();
   void recordRawValue(long duration);
}
```

* Anything that needs to provide profiling information implements the `Profileable` interface

```Java
interface Profileable {
    void setProfiler(Profiler profiler);
}
```

Each metric of interest gets a dedicated measure. The field initialize to a `DISABLED` implementation with empty methods that should be able to be removed by dead code elimination. 

```Java
class SpecialAxis implements Profileable {

    public void layoutChildren() {
        benchLayoutChildren .start();
        // do work...
        benchLayoutChildren.stop();
    }

    public void drawAxis() {
        benchDrawAxis.start();
        // do work...
        benchDrawAxis.stop();
    }

    @Override
    public void setProfiler(Profiler profiler) {
        benchLayoutChildren = profiler.newDuration("specialAxis-layoutChildren");
        benchDrawAxis = profiler.newDuration("specialAxis-drawAxis");
    }

    private DurationMeasure benchLayoutChildren = DurationMeasure.DISABLED;
    private DurationMeasurem benchDrawAxis = DurationMeasure.DISABLED;

}
```

### Debug printer

The simplest profiler is a debug printer that prints start/stop tags and some timing information to a log.

```Java
chart.setProfiler(Profiler.printProfiler(System.out::println, boolean printStartedInfo));
```

```text
chart-runPreLayout - started
chart-lockDataSets - started
chart-lockDataSets - finished (0.01 ms)
chart-updateAxisRange - started
chart-updateAxisRange - finished (0.39 ms)
chart-runPreLayout - finished (0.53 ms)
chart-cssAndLayout - started
chart-layoutChildren - started
chart-layoutChildren - finished (0.34 ms)
chart-cssAndLayout - finished (0.57 ms)
chart-runPostLayout - started
chart-drawAxes - started
chart-drawAxes - finished (0.09 ms)
chart-drawCanvas - started
chart-drawCanvas - finished (1.16 ms)
chart-runPostLayout - finished (1.36 ms)
```

### Tag selection

Via default methods we can create a fluent API for filtering tags of interest and modify the tags with meta-information for better disambiguation.

```Java
var profiler = Profiler.debugPrinter(System.out::println);
chart.getXAxis().setProfiler(profiler.addPrefix("x-").contains("computePrefSize"));
chart.getYAxis().setProfiler(profiler.addPrefix("y-").contains("computePrefSize"));
```

```text
y-axis-computePrefSize - started
y-axis-computePrefSize - finished (8.85 ms)
x-axis-computePrefSize - started
x-axis-computePrefSize - finished (5.67 ms)
y-axis-computePrefSize - started
y-axis-computePrefSize - finished (0.02 ms)
x-axis-computePrefSize - started
x-axis-computePrefSize - finished (0.01 ms)
```

### HdrHistogram output

[HdrHistogram](http://hdrhistogram.org/) is an established and very low-overhead way to measure production behavior in low-latency environments (banking, real-time systems etc.). It records latencies in high-dynamic range histograms (i.e. resolution depends on the absolute value, so it's very space efficient) and provides a file format that can be read by various tools. I've used it a few times, so I had some code sitting around.

The profiler records histograms and continuously writes them to disk in set intervals. The histograms are stored with a tag so they can be disambiguated again later. Profiling an axis and loading it in [HdrHistogramVisualizer](https://github.com/ennerf/HdrHistogramVisualizer) (a tool I wrote a long time ago when testing the normal JavaFX Charts) looks as below

```Java
chart.getXAxis().setProfiler(Profiler.hdrHistogramProfiler("bench/axis.hlog"));
```

The top chart shows the maxima for each interval, and the lower chart shows a percentile plot with an exponential x axis.

image

### Live chart

Given that this is a charting library, I figured it'd be appropriate to also have a live chart of the real-time generated profiling data:

```Java
var profiler = Profiler.chartProfiler("RollingBufferSample");
chart.setProfiler(profiler.matches(".*[run|draw].*"));
chart.getXAxis().setProfiler(profiler.addPrefix("x-").contains("draw"));
chart.getYAxis().setProfiler(profiler.addPrefix("y-").contains("draw"));
beamIntensityRenderer.setProfiler(profiler.addPrefix("beamIntensity-").contains("render"));
dipoleCurrentRenderer.setProfiler(profiler.addPrefix("dipoleCurrent-").contains("render"));
```

I already had some code for HdrHistogram-like visualizations, so I put together an initial demo of it profiling itself. [Here is a video](https://www.youtube.com/watch?v=n75JJF5NTvQ).

image


ennerf:
After review I think that the measurements are benchmarks rather than profiling info, so I changed the names to reflect that. The PR also adds a `BasicDataSetRenderer` that is more efficient for rendering standard datasets without errors (based on a renderer we use internally).

Example 

```Java
var recorder = MeasurementRecorders.newLiveDisplay("RollingBufferSample").matches(".*(all).*");
beamIntensityRenderer.setRecorder(recorder.contains("all").addPostfix("error"));
```

Issue #605, opened 2023-08-24
DragResizerUtil dont resize on south


rsuna:
**Describe the bug**
Its not possible to resize a region by south, is a small copy&paste mistake. PR is pending.

**To Reproduce**
Try to resize a component at the bottom is not working.

**Environment:**
 - OS: WIndows10
 - Java version: 11
 - JavaFx version:  11
 - ChartFx version: 11.2.2

RalphSteinhagen:
closing as fixed by already merged PR #606

PR #606, opened 2023-08-24
fix DragResizerUtil for south

rsuna:
Make south resize working. See issue #605.


wirew0rm (dismissed):
Looks good, Thanks for catching this and directly providing the fix!

I'll apply the formatting patches and merge.

PR #614, opened 2023-09-21
benchmark module

ennerf:
Simple PR that moves the benchmark classes to a dedicated module that is compatible with jdk8.

This allows it to be used in older code bases that can't depend on dataset (jdk11).


wirew0rm (approved):
Looks good, as discussed.

PR #615, opened 2023-09-21
Simplify extending the ErrorDataSetRenderer

### LINKED ISSUE #613: Change CachedDataPoints access modifier to public
**Describe the bug**
Classes that extend the ErrorDataSetRenderer and override any of the `draw*(GraphicsContext, CachedDataPoints): void` methods must be placed in the same package (`de.gsi.chart.renderer.spi` or `io.fair_acc.chartfx.renderer.spi`) because CachedDataPoints has only package private class access and is exposed outside of its defined visibility scope.
If there are no objections, we should change the access modifier to public. This will make it easier to customize the error data set renderer and to avoid split packages.
Let me know if you are interested in a PR for this small change.

**Environment:**
 - ChartFx version: main/11.2.x

TheBoegl:
This commit changes the access modifier for the CachedDataPoints class to public and fixes #613. It also makes it easier to customize the error data set renderer and avoids split packages.

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/CachedDataPoints.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/CachedDataPoints.java
@@ -24,13 +24,13 @@ import io.fair_acc.dataset.utils.ProcessingProfiler;
 import io.fair_acc.math.ArrayUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


wirew0rm (approved):
Looks good, thanks for contributing back. Changing this on the main branch isn't a problem, i assumed first that you wanted to target this to the released version, which would have been more complicated since we're not really planning on maintaining multiple versions.

PR #618, opened 2023-09-25
BenchPlugin

ennerf:
Added an experimental `BenchPlugin` that displays a live benchmark of the current chart:

image

Clicking again or closing the window disables the measurements again

image


wirew0rm (dismissed):
Looks good!

Adds an easy way for library users to use the basic features of the instrumentation and serves as a nice showcase in the samples.

PR #620, opened 2023-09-26
minor fixes

ennerf:
PR with some tiny fixes:

* Made list of BitState listeners thread-safe
* Changed `BenchPlugin` to only allow a single open Window (fixes reset when a 2nd window is closed)
* Limited default bench measures to the most interesting ones
* Changed `SimplePerformanceMeter` to use the `Node::isDirtyEmpty` method instead of the `Node::dirtyBits` field which was changed/removed at  one point.


wirew0rm (approved):
:+1:

PR #623, opened 2023-09-26
made BenchPlugin more extensible to support custom CSS files

ennerf:
I missed an extension for adding CSS styling to the benchmark chart


wirew0rm (approved):
:+1:

PR #624, opened 2023-09-28
changed label setting to match 11.2.7 behavior

ennerf:
I double checked the automatic axis-label setting, and the 11.2.7 behavior was that each renderer tries to use the first dataset. Once the labels get set, a boolean state ensures that the labels do not get overwritten later.

This PR changes the behavior to match the 11.2.7 release.


wirew0rm (approved):
thanks, looks good :+1:

PR #628, opened 2023-09-29
RotatedAxisLabelSample: Disable AutoRanging

wirew0rm:
Having autoranging enabled on synchronized axes with bound min/max properties results in errors when the bound properties are modified by the autoraning algorithm.


ennerf (approved):
going forward the `min/max/tickUnit` properties should be ReadOnly similar to the `scale` and `length` parameters.

PR #631, opened 2023-10-09
Zoomer: add a user definable predicate for panning

ichaus-we:
This commit adds a user-definable predicate for panning to the Zoomer plugin.

It allows to replace the default predicate (middle mouse button) to something else, like it was possible with the now removed Panner plugin.

Sample code:
`zoomer.setPanMouseFilter(event -> MouseEventsHelper.isOnlyPrimaryButtonDown(event) && MouseEventsHelper.isOnlyCtrlModifierDown(event));
`

This is demonstrated in a new chart in the ZoomerSample.


wirew0rm (approved):
Thanks for this nice improvement, looks good!

From my pov the now unused predicate could also have been directly removed instead of deprecated, but we can also do that after the next release.

PR #669, opened 2024-05-25
Fix gradle artifactId in README

pacher:
Example gradle snippet still uses old `chartfx-chart` artifact id, which does not exist in the new `io.fair-acc` group.

Also updated the versions to the latest 11.3.1


wirew0rm (approved):
Thanks for the fix, this must have been overlooked with the last releases. Hope it didn't cause too much confusion.

PR #671, opened 2024-06-26
DefaultAxisTransform: fix autoscale rounding direction

protogenes:
DefaultAxisTransform is using the wrong rounding operation which leads to cut off range or missing labelling


RalphSteinhagen (approved):
@protogenes thanks for spotting and directly posting a PR! Much appreciated!

PR #672, opened 2024-08-15
Fix update of line dashes

protogenes:
Changing `Number[] lineDashArrayProperty` will not update the binding `double[] lineDashesProperty` which is used by the renderer (via `StyleUtils`), thus any change will never be shown on screen.

This is just a quick fix, in my opinion it would be better to remove `lineDashesProperty` and use some variant of `toLineDashArray` to convert the `Number[]` when rendering


wirew0rm (approved):
Thanks for catching this, thanks for the explanation and the fix :pray: :+1: 

I'll have to look into why we have both the Number[] and double[] properties there and whether it is still needed or can be handled during the rendering as you propose. For now I would assume that since these should be small arrays that rarely change the overhead for the binding shouldn't cause too much trouble but of course it makes following the code less straightforward, so if it's possible to simplify that we should probably do that.

PR #673, opened 2024-08-19
Chart: fix binding disposal when removing plugins

protogenes:
fix parameters to `Bindings.unbindContent` to actually dispose the binding created for the plugin group


wirew0rm (approved):
lgtm :+1:

PR #676, opened 2024-09-23
ChartPlugin scene listener is not removed

amischler:
The `ChartPlugin` register a listener on some of its nodes `sceneProperty` but those listeners are never being removed, even when the plugin is removed from a chart (i.e. `chart` property of the Plugin is set to `null`).

In some cases, a `ChartPlugin` that is not in any chart might still receive `Scene` events, which should not be the case.

See `ChartPluginTest` in this PR for more details.


wirew0rm (approved):
Thanks for finding, fixing and contributing back :+1:  This looks like a very comprehensible fix, well documented and with unit tests, thanks a lot and sorry for taking so long to look at it.

amischler:
Thank you for merging and BTW thank you for this great library!

PR #678, opened 2024-11-20
BP: AbstractRendererXY: Update axesList of AbstractRenderer when changed (#677)

### LINKED ISSUE #677: Plugins like DataPointTooltip are not working at all if axes were only added to the chart
**Describe the bug:**

Plugins like `DataPointTooltip` are not working at all if the _axes were only added to the chart_ and not also to the renderer.

**To Reproduce:**

Open `FxmlSample` and hover an a data point. No tooltip appears although the plugin is active in this sample.

Screenshot at 2024-11-20 13-37-18

**Expected:**

Screenshot at 2024-11-20 13-37-55

**Environment:**
 - OS: Debian 12.7
 - Java version: openjdk version "17.0.13" 2024-10-15
 - JavaFx version: 16
 - ChartFx version: main, 05980c41

**Fix:**

I have a fix prepared for `AbstractRendererXY` and will create a PR for it.

dedeibel:
Fixes #677

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/AbstractRendererXY.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/AbstractRendererXY.java
@@ -74,6 +74,9 @@ public abstract class AbstractRendererXY<R extends AbstractRendererXY<R>> extend
 
     @Override
     public void updateAxes() {
+        final var xAxisBefore = xAxis;
+        final var yAxisBefore = yAxis;
+
         // Default to explicitly set axes
         xAxis = ensureAxisInChart(getFirstAxis(Orientation.HORIZONTAL));
         yAxis = ensureAxisInChart(getFirstAxis(Orientation.VERTICAL));
@@ -94,6 +97,10 @@ public abstract class AbstractRendererXY<R extends AbstractRendererXY<R>> extend
         }
         if (yAxis instanceof CategoryAxis axis && !getDatasets().isEmpty()) {
             axis.updateCategories(getDatasets().get(0));
+        }
+
+        if (xAxisBefore != xAxis || yAxisBefore != yAxis) {
+            getAxes().setAll(xAxis, yAxis);
         }
     }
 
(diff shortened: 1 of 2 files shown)


chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/AbstractRendererXY.java
```diff
@@ -95,6 +98,10 @@ public void updateAxes() {
         if (yAxis instanceof CategoryAxis axis && !getDatasets().isEmpty()) {
             axis.updateCategories(getDatasets().get(0));
         }
+
+        if (xAxisBefore != xAxis || yAxisBefore != yAxis) {
```
RalphSteinhagen, line 102: This may benefit from a check whether the original [x,y]Axis isn't null and decouple the setAll(..) call. 

Otherwise OK.

RalphSteinhagen (approved):
@dedeibel thanks for your PR! This looks OK. Thanks for also providing a unit-test.

As a minor request: if you could add a null pointer check for the rare case that there isn't any original axis before the updateAxes() call. We do not want to nullify any valid new axis. :smile: 

Since this is minor, it could be squash-merged afterwards w/o further review.

chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/AbstractRendererXY.java
dedeibel: Thank you for looking into it. I am afraid I am unsure what you mean. Could you provide a code snippet that I could use?

dedeibel:
Thank you for your review! I will keep your suggestion in mind and hope we can get back to it if any problems with null values arise.

chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/AbstractRendererXY.java
RalphSteinhagen: ... probably a rare (or even impossible) corner case. Will merge this and -- in the unlikely event that this causes problems -- will follow this up.

PR #692, opened 2025-05-11
Make it compile in Ubuntu.

lacgit:
When "mvn compile install", it throw "java.lang.OutOfMemoryError: Java heap space: failed reallocation of scalar replaced objects".

Changing -Xmx2048m to -Xmx4096m in below of the pom.xml fixes this error.

            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-surefire-plugin</artifactId>
                <version>${chartfx.surefire.version}</version>
                <configuration>
                    
                    <argLine>${argLine} -Duser.language=en -Duser.country=US -Xms256m -Xmx4096m -XX:G1HeapRegionSize=32m -Djava.awt.headless=true -Dtestfx.robot=glass -Dtestfx.headless=true -Dprism.order=sw</argLine>
                    <forkCount>1</forkCount>
                    <rerunFailingTestsCount>3</rerunFailingTestsCount>
                    <runOrder>random</runOrder>
                </configuration>
            </plugin>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -254,7 +254,7 @@
                 <version>${chartfx.surefire.version}</version>
                 <configuration>
                     <!-- Jacoco prepare-agent builds some command-line params without which jacoco will not instrument. Hence it is important to add those command-line params here (${argLine} holds those params) -->
-                    <argLine>${argLine} -Duser.language=en -Duser.country=US -Xms256m -Xmx2048m -XX:G1HeapRegionSize=32m -Djava.awt.headless=true -Dtestfx.robot=glass -Dtestfx.headless=true -Dprism.order=sw</argLine>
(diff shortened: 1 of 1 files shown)


wirew0rm (approved):
Thank you. I don't think this is really ubuntu specific, especially since the CI is also running on ubuntu, but some of the allocation strategies might have changed causing memory to be reclaimed later and I don't see much harm in increasing this limit.

For reference, the `-Xmx` flag sets the maximum heap-size of the jvm. The change in this commit increase this limit for the unittest-runner from 2 gigabyte to 4 gigabyte, after reports of the testsuite failing with out-of-memory on certain systems.

It would probably also be interesting to investigate a memdump of when this occurs to check if we cannot otherwise reduce the resource consumption.

PR #701, opened 2025-07-25
Control on/off to draw tick mark label while keep the tick mark visible

lacgit:
Allow user to control turn on/off the tick mark label while keeping the tick mark visible.
 This can be useful if we would like to stack up a series of small charts with the same x-axis
 This hope to get the UI tighter.


ennerf:
While maybe  a good idea, the parameter would need to be accessible via CSS, invalidate appropriate parts of the chart on change, and be included in the layout/bounds computations.

That being said, I believe that you should already be able to do this via CSS and programmatically

```Java
chart.getXAxis().getTickLabelStyle().setVisible(false);
```

```css
.axis .axis-tick-label {
  visibility: hidden;
}
```

If that doesn't do what you want, we should do a bugfix.

lacgit:
Thanks Florian,
I have tried that but it disable the tick mark as well, which I would like to keep to give some relativity to the viewer.  May be I can enhance this to include the CSS accessibility and the other change invalidation handling too.
Regards,

PR #702, opened 2025-07-26
fixed a bug that caused tick marks to not render without labels

ennerf:
See #701 

```css
.axis .axis-tick-label {
  visibility: hidden;
}
```

image

```css
.axis .axis-major-tick-mark, .axis .axis-minor-tick-mark {
  visibility: hidden;
}

.axis .axis-tick-label {
  visibility: visible;
}
```
image


RalphSteinhagen (approved):

RalphSteinhagen:
Thanks @ennerf for the fix. Looks good and seems to do the job (I assume you've tested more thoroughly than I can 😃).

ennerf:
It looks like there is a problem with the CI checks

Issue #695, opened 2025-06-12
Bars get rendered with infinite width when errorType != ERRORCOMBO


ennerf:
**Describe the bug**

Bars render correctly with the default error type (`ERRORCOMBO`), but they render across the whole width under any other setting.

This is because `ERRORCOMBO` calls `points.getMinXDistance()`, which does an on the fly calculation that sets the `minDistanceX` field.

```Java
case ERRORCOMBO:
    if (points.getMinXDistance() >= getDashSize() * 2) {
        drawErrorBars(gc, style, points);
    } else {
        if (isallowNaNs()) {
            drawErrorSurfaceNaNCompatible(gc, style, points);
        } else {
            drawErrorSurface(gc, style, points);
        }
    }
```

The [ErrorDataSetRenderer.java:L169](https://github.com/fair-acc/chart-fx/blob/8878b813e21effdb37f5618cc061c80ad71420a0/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/ErrorDataSetRenderer.java#L169C69-L169C81) then accesses the set field directly:
```Java
final int minRequiredWidth = Math.max(getDashSize(), points.minDistanceX);
```

![Image](https://github.com/user-attachments/assets/96e11712-bc22-48b7-a743-a0cddfe54012)

Since the other error modes never set it, the minDistanceX is always `Integer.MAX_VALUE` by [CachedDataPoints:L51](https://github.com/fair-acc/chart-fx/blob/8878b813e21effdb37f5618cc061c80ad71420a0/chartfx-chart/src/main/java/io/fair_acc/chartfx/renderer/spi/CachedDataPoints.java#L51) and all bars render across the whole screen.

![Image](https://github.com/user-attachments/assets/704b45a9-2b9d-4f9b-a170-aa02fc000936)

It should be fixed by replacing the raw field access with the method, but I haven't had a chance to confirm it yet.

```Java
final int minRequiredWidth = Math.max(getDashSize(), points.getMinXDistance());
```

wirew0rm:
Looks right, thanks for the report and the detailed analysis. Since the first statement in the getter is to do an early return with the cached value anyway and this is not in a tight loop replacing this looks like the right way to fix it.

ennerf:
I looked a bit more into this, and the point reducer also resets and recomputes the value.

So this problem comes about when

```Java
final ErrorDataSetRenderer renderer = new ErrorDataSetRenderer();
[...]

PR #703, opened 2025-07-26
fixed bar rendering issue #695

ennerf:


wirew0rm (approved):
Thanks for the changes and the [detailed investigation](https://github.com/fair-acc/chart-fx/issues/695#issuecomment-3122179638). The changes look very nice and concise.

PR #704, opened 2025-08-13
BP: EditAxis plugin: prevent OOM when adding and removing axes dynamically

dedeibel:
I noticed this problem of my program going OOM when I implemented a feature that involved adding and removing axes dynamically.

Please accept my PR or adapt it as needed.

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/plugins/EditAxis.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/plugins/EditAxis.java
@@ -13,6 +13,7 @@ import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.value.ChangeListener;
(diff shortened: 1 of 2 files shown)


chartfx-chart/src/main/java/io/fair_acc/chartfx/plugins/EditAxis.java
```diff
@@ -128,7 +132,11 @@ protected void addMouseEventHandlers(final Chart newChart) {
             return;
         }
         newChart.getAxes().forEach(axis -> popUpList.add(new MyPopOver(axis, axis.getSide().isHorizontal())));
-        newChart.getAxes().addListener(this::axesChangedHandler);
+
+        /* This method is potentially called multiple times, make sure listener is only present once (contains check
+         * is not available) */
+        newChart.getAxes().removeListener(axisChangeListener);
```
dedeibel, line 138: This is the relevant line.

wirew0rm (approved):
Thanks for the fix and the detailed description and tests, looks good :+1:

PR #706, opened 2025-11-14
Bugfix/705 zoomer plugin issues with inverted axes

### LINKED ISSUE #705: Zoomer plugin issues with inverted axes
**Description**
I try to use the zoomer plugin with an inverted axis. But when zooming (with the mouse wheel), the axis range flickers or jumps, and for some zoom settings the axis labels vanish completely. The displayed data then either jumps according to the axis, or sometimes seems to disappear completely.

**Reproduction**
This can be reproduced with the ZoomerSample. I simply changed the method `getTestChart` to
```
private static Chart getTestChart(final String title, final DataSet testDataSet) {
        final Axis xAxis = new DefaultNumericAxis();
        xAxis.invertAxis(true);
        final Axis yAxis = new DefaultNumericAxis();
        yAxis.invertAxis(true);
        final var chart = new XYChart(xAxis, yAxis);
        chart.setTitle(title);
        chart.setLegendVisible(false);
        chart.getDatasets().add(testDataSet);
        chart.setPrefSize(PREF_WIDTH, PREF_HEIGHT);

        return chart;
    }
```
to invert the axes (which for some reason only worked for the y axes, but that was sufficient for me). Then run the sample and zoom, and the bug should be obvious.

**Environment:**
 - OS: Rocky Linux 9
 - Java version: 17
 - JavaFx version: 16
 - ChartFx version: 11.3.1

dedeibel:
Fixes #705 and related problems.

@wirew0rm could you please review and accept these changes.

### DIFF
--- a/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/DefaultNumericAxis.java
+++ b/chartfx-chart/src/main/java/io/fair_acc/chartfx/axes/spi/DefaultNumericAxis.java
@@ -604,8 +604,6 @@ public class DefaultNumericAxis extends AbstractAxis implements Axis {
 
             logScaleLengthInv = axisLength / logScaleLength;
             offset = axisLength;
-
-            offset = isVerticalAxis ? getHeight() : getWidth();
         }
     }
 }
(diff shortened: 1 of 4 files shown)


dedeibel:
Is there an estimate on when you might get to looking at the PR?

RalphSteinhagen (approved):

